PR #7, opened 2018-06-19
Fix flag name in lint command for Travis CI

kytrinyx:
I missed this when bootstrapping.


kytrinyx:
I'm going to go ahead and merge this, since it should have been done this way in the first place!

PR #8, opened 2018-06-19
Add auto_approve to hello-world exercise

jackhughesweb:
This adds the auto_approve property to the hello-world 
exercise as part of preparing tracks for the v2 launch.

The current code on the server will automatically approve
hello-world exercises anyway, but that is a temporary fix
and will be removed in the future.

See exercism/pharo#5


macta:
@jackhughesweb I can't see an auto-approve property in that config file? Is it just by having a config file that you get auto-approve? If so, we now have a config file and so maybe I can close this?

OR do I need to add something explicitly to the file we now have?

Issue #10, opened 2018-07-09
Write first few exercises


samWson:
We need a few exercises to get this language track started.

I haven't seen any activity in this repo for a few days. I thought I would get things moving by making some of the first few exercises, an also make sure I have the repo structure correct according to [this discussion](https://github.com/exercism/exercism-pharo-tooling/issues/4#issuecomment-402575116).

First pull request to follow soon. After that we can have some discussion to make sure the structure of the repo is on the right track.

#### Todo
- [x] HelloWorld exercise
- [x] TwoFer exercise

samWson:
I've made a pull request for this at #11. Please read, review, and comment.

bencoman:
To avoid confounding factors, I converted PR #11 to Tonel format in the 'master' branch of my own repo.
@samWson, @macta could you try it out in latest Pharo 7 like this...
![clone-pharo-exercism-tonel](https://user-images.githubusercontent.com/1713447/42694399-e1c8ce9a-86e4-11e8-8dd2-c41927cff679.png)

I'm not entirely satisfied with my structure, but I'd like your fresh view on it before adding my own.

Also I discovered that having the exercism repo called "pharo" conflicts with the default "pharo" repo definition built into Pharo 7.

samWson:
@bencoman I'm using PharoLauncher on MacOS. I just downloaded a new image of Pharo 7 last night. I've copied and pasted the exercise code into the Pharo 7 image and made a new git repo. 

> Also I discovered that having the exercism repo called "pharo" conflicts with the default "pharo" repo definition built into Pharo 7.

I've managed to name the repo `pharo` without any problems so far. Although I create new git repositories by using the command line tool and then using Iceberg to add a local repository (I've had a miserable time getting Iceberg to make new repos). However I think it would be a good idea to make a better name to avoid potential conflicts.

I'll push the latest version of my work to the PR later tonight.

macta:
Guys - I’d target Pharo 6, as 7 is going to be months (maybe even early next year) before it’s ready.

I’d hope we could have exercism launched next month with a few exercises?

If we do choose 7, the ide extension is a bit different (I actually wrote it first but then concluded 6.1 might be a better stable target)

samWson:
The only barrier to using Pharo 6 is just getting started with Tonel in Iceberg as far as I can tell. From the Tonel [repository]() it looks simple enough to install Tonel, and as long as you have the `.properties` file specifying Tonel format Iceberg should use it correctly.

[...]

PR #11, opened 2018-07-11
WIP: Issue 10 - Write first few exercises

samWson:
Solution for #10.

I've just written the first two exercises: HelloWorld and TwoFer.

@bencoman @macta I've tried to structure things according to our discussion [here](https://github.com/exercism/exercism-pharo-tooling/issues/4#issuecomment-402575116). My understanding is that the `ExercismDev` package is where exercises under development or reference solutions are placed. These two exercises could be considered reference solutions in that case. This means that the `ExercismDev` package is not for students but for development of the language track.

As I understand it the rest of the repo is supposed to work as follows: in the package `ExercismDevTools` an `ExercismGenerator` will be used to take the passing tests from the two exercises here and produce the Tonel files of each, placing them in the `Exercism` package. These files are what will be downloaded to the students image and are the tests for which a solution will be developed. A fourth package, `ExercismTools` handles pushing and fetching solutions.

That's how this repo is to work as a whole, as I understand it. Let me know if I have anything wrong.

I've left my code in the default format that Iceberg produces. I don't know if anyone wants to use Tonel for that too, or just for the files we want to be human readable. I don't know how to change that format myself.

Please read, review, and comment.


macta:
@samWson This work needs to be redone as its using an older Tonel format (I made the same mistake in my spike). I've submitted a PR to tonel to document this better but essentially you need to create a .properties file under /src that has
```
{
	#format : #tonel
}
```

The older default Filetree format (with one file per method) is too hard to review in github. So lets start out how we mean to go.

Also - we still seem to have differing ideas on the structure of the project. I think its too much hassle to have separate projects for each exercise (i.e. having /excercises/HelloWorld/src).

I would organise it as /exercises/HelloWorld/*.st, /exercises/TwoFer/*.st etc - these paths and contents are generated by /src/ExecercismDev/ so it will require two checkins - one via Iceberg and a second one when we run the Generator to create the exercises that students will see.

macta:
With regards to TwoFer, I think we should perhaps be more pedantic with naming - tf is not a great variable name? Also should discuss the convention of instance vs class methods on simple exercises. At what point do you swap from class to instance with education in mind? I’m going to pop a question on the Pharo group to see what others think

bencoman:
> older Tonel format

I'm not certain on this, but don't think its an older "Tonel" format, just an older format called "filetree" that initially Iceberg used.  However this results in a large number of small files which had performance problems on Windows and also had trouble with file path length limits with long method names.

> With regards to TwoFer, I think we should perhaps be more pedantic with naming - tf is not a great variable name
I'm in the process of updating the examples to use Tonel format from Pharo 7,  
but in the meantime I'd generally agree with @macta on using full variable names, 
but with such a simple example...   
     self assert: (twofer twoFer: 'Alice') equals: 'One for Alice, one for me.'. 
seems a bit awkward. 
So for that one I'm leaning towards...   
    self assert: (testobject twoFer: 'Alice') equals: 'One for Alice, one for me.'.

samWson:
I'm more than happy to use Tonel as the format for everything in Iceberg. It's just a case of me not knowing how to get Iceberg set up correctly. @macta if it is as simple as a `.properties` file I'll give it a try, however would that still work on my already existing repository or will I need to junk it and start again? At this early stage it is not much trouble to start from scratch if we need to.

@macta is this directory structure what you are describing?
```
pharo
  |--exercises
  |      |--HelloWorld
  |      |      |--*.st
  |      |--TwoFer
  |             |--*.st
  |--src
      |--ExercismDev
```
For example the exercises in this pull request would go in `ExercismDev`, and from those exercises the files that students would see are generated and placed in `exercises/HelloWorld` and `exercises/TwoFer`. Is this correct?

@macta @bencoman I agree using `tf` is a bad idea. I couldn't come up with a solution that read well so I pushed it here for discussion. In fact I don't like how these one method solutions read at all on simple exercises. Emphasizing the Smalltalk convention of readable code would probably be a good idea early on. So I wonder if the better approach is to use a class method and have the solution like `TwoFer who: 'Alice'`. Otherwise `testObject twoFer: 'Alice'` is better. I've seen `subject` or `described_class` used in similar ways.

Thanks for the suggestions. This discussion is exactly whats needed. I'm happy to keep iterating on this pull request. Unless you want me to do otherwise, the next time I push I will `--force` it, overwriting the work shown here since it will need to change the entire format.

bencoman:
@samWson I manually imported your examples to review them.  I came up with an idea of putting the solutions as an extensions.  I've attached the fileouts for review while I dig into an issue I've got with Iceberg.
[Exercism-Issue11-first-few-exercises.zip](https://github.com/exercism/pharo/files/2187873/Exercism-Issue11-first-few-exercises.zip)

macta:
@samWson yes that structure is what I was thinking. Iceberg will manage src and below (it’s currently not capable of doing more) - but this should be fine. And yes  the exercises and solutions that we develop will also live in src in Tonel format, but in a package only devs will load, but we will export (generate) the relevant bits into that exercism directory for students to run with (we will need to use cmd line git to do commit that).

Any word on us getting commit access too so we can test it out with exorcism itself?

For converting to tonel2 I would blow it away and redo it. There is a conversion tool but it changed recently - I haven’t tried the new way yet.

We now need to write the ExercismGenerator which should be pretty easy.

macta:
@bencoman solutions as extensions isnt a bad idea - although will that work if there is no solution class defined? I think we should leave out the solution class and emphasise that in Pharo you press create in the debugger which was the real TDD intent. I notice that the ruby examples also take this approach and talk about tdd (they just don’t have such good tools as us)

samWson:
>Any word on us getting commit access too so we can test it out with exorcism itself?

@kytrinyx are you able to help with this? I don't have access to the `exercism/pharo` settings (the icon isn't visible) so I can't add @macta and @bencoman as collaborators to this repo.

Until we can get you two collaborator access I'll get an issue for the ExercismGenerator posted for you. 

I'll keep working on this issue for the time being, after I sleep. It's late in my part of the world. Thanks for the help team.

samWson:
I've pushed the exercises again in Tonel format this time with a few minor changes. The pull request has closed by itself however. I think this is be cause I forced push over the original commits with code from a completely new repo. So they have no common history. You can still view the commits as whole files rather than diffs.

I might just make another pull request called `issue-10-v2`.

PR #20, opened 2018-07-25
Update pharo track intro text and provide snippet

macta:


macta:
As this is just some documentation files, I'm going to merge it

PR #23, opened 2018-07-26
Improved installation readme

macta:
Small change as I tested loading from the office and hit a small snag. Also tried to make it read and look a bit better


samWson:
Looks good to me. Go ahead and merge if you are happy. Note for later though we will have to test that the installation and image starting steps work for the major operating systems. I've got MacOS, Windows 10, and Ubuntu Linux available so this shouldn't be a major hurdle.

macta:
Cool - you've got the full set to test it with. I'm going to rename the hello-world example so that @kytrinyx can get us back to beta status which should mean we can do low level testing to prove the idea works (I will need to do to more work in the generator to match their conventions)

Issue #24, opened 2018-07-26
Rename `HelloWorld` directory to `hello-world`


kytrinyx:
In order to be able to add the Pharo track to the site, we need the Hello, World exercise directory to be named `hello-world`. The files within it can be named per the naming conventions of the language.

See more about the anatomy of exercises in https://github.com/exercism/docs/blob/master/language-tracks/exercises/README.md

kytrinyx:
I'll need to update the instructions, in that case.

The config slugs must be all lowercase, and hyphenated. The exercise slug is used as an identifier by the website, as well as to identify the exercise by the command-line app. Additionally, we use the slug to connect the exercise implementation in each language to the problem specification for the same exercise in https://github.com/exercism/problem-specifications.

The files themselves, or subdirectories within it could all have camel-case, but we have a lot of things that rely on the underlying assumption that the directory itself is normalized to kebab-case.

macta:
@kytrinyx ok - this where the onboarding for developers of a track is all a bit mysterious - the docs are a bit sprawling and its not obvious what is configuration and what is convention (particularly for non c-style languages that have different conventions). In retrospect it makes sense now - I'll just need to do a bit more work in our mapping-generation.

I'm hoping this is all worth while, as it seems a shame not to have languages that work a bit differently in your tracks - and give people an opportunity to learn something a bit different (I guess with common tools like exercism that they feel comfortably with).

I think we can pull it off.

macta:
This has been committed

kytrinyx:
> I'm hoping this is all worth while, as it seems a shame not to have languages that work a bit differently in your tracks

Yes! I definitely want to be able to support languages that work differently, it's just not come up before. I think this means you're a pioneer :)

PR #26, opened 2018-07-26
Natural domain class naming

macta:
This might relate to #24 if there are other less obvious naming conventions going on


macta:
rats - Go doesn't support negative lookahead

PR #27, opened 2018-07-26
Improve lint checker regex (Go compatible), so it fixes #25

### LINKED ISSUE #25: Rework config.json Test and domain class lint checking
The exercism config.json has entries to match test classes, solution classes and ignore classes.

It defaults to regex to match [Tt]est for tests and [Ex]xample for domain classes. The future generator could write files out to fit this, but @samWson pointed out it feels weird to write tests that instantiate and DomainExample class (e.g. HelloWorldExample new) - we should be able to just write a better regex for such a simple lint check.

macta:
As per comments from @samWson about domain class naming in tests and not wanting to make our generator too complicated

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,8 +3,8 @@
   "language": "Pharo",
   "active": false,
   "blurb": "Pharo is a pure open source object-oriented programming language and a powerful environment, focused on simplicity and immediate feedback (think IDE and OS rolled into one).  It has an elegant syntax that is natural to read and write.",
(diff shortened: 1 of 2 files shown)


macta:
I'm going to merge this one - we may need to tweak it more when we get back into beta status and fit more conventions

samWson:
> I'm going to merge this one - we may need to tweak it more when we get back into beta status and fit more conventions

I predict we are going to do a lot of rough work that we will need to fix later in these early stages. It's fine to merge things now. We can make changes after we try things out.

PR #28, opened 2018-07-27
rename HelloWorld to match exercism kebab-case requirement (also fixe…

macta:
…d config.json to match along with a configlet generated uuid)


macta:
I'm going to merge this - so hopefully @kytrinyx can get us back into beta status so we can try some things out and understand how we make this work seemlessly

PR #29, opened 2018-07-27
Create initial generator

macta:
@samWson This should give us the basic capabilities we need. It will need refining, and I haven't been able to run it using Exercism V2 (so its guess work) but it looks to write out the correct bits, and I can manually cause it to read in a written out kebab fmt directory.

You should then be able to put your example HelloWorld and TwoFer into the Exercism package (as sub-projects named in camel case)and we could then ExercismGenerator generate them. (will need to work on the instructions for this though)


macta:
#fixes #29

PR #34, opened 2018-07-29
Fairly comprehensive .gitIgnore to help less advanced contributors (g…

### LINKED ISSUE #33: Improve .gitIgnore for multi lang development
As Pharo doesn't have full git support yet - many contributors will use an external VCS tool like Idea or VCS - so improve the .gitIgnore so we don't end up with extra junk in the repo

macta:
…enerated via IntelliJ gitIgnore.io)

fixes #33 

@samWson I think we are thinking the same thing here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,200 @@
+### Exercism default
 *.swp
-.DS_Store
 bin/configlet
 bin/configlet.exe
+
+### Smalltalk template
+# changes file
+*.changes
+
+# system image
+*.image
+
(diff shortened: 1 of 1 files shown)


samWson:
Overkill perhaps but it doesn't hurt. We can always add or remove later.

macta:
Yeah - I think it can be trimmed down. I noticed it when using IntelliJ to help checkin some of the other .md docs (and there were a bunch of distracting misc files).

I didn't want to mess with it too much - I think we need to use it and see and then clean it up.

PR #35, opened 2018-07-30
Adds an all tests suite to easily run all associated project tests (w…

macta:
…ill need this for CI)

First steps to CI (but also handy as a user to click on the all tests orb to run everything in the IDE - we might consider adding an all tests in the Exercises project for users - but didn't want to confuse things for now)


samWson:
Also handy for devs. The test runner could also do it too, but this would be just that extra bit convenient for testing everything Exercism.

Issue #18, opened 2018-07-24
Update the Pharo track intro text and provide a sample snippet


macta:
When you first view the Pharo track, you are presented wit a concise summary of the language and why its important as well as simple code snippet.

We need to update the /docs/about.md and /docs/snippet.txt to wow people about Pharo and importance of Smalltalk

kytrinyx:
Once the directory name is sorted out I can add Pharo back to to the website in beta mode.

macta:
we’re waiting for @kytrinyx to come back from hols (Aug/1) to enable beta. Not sure if this will let us see the website page and logo in-situ to tweak it, but should let us try fetching examples from inside Pharo and submitting solutions and possibly seeing what mentor commenting is like?

macta:
This is now unblocked, and the there is PR under review to tweak it.

macta:
The initial docs are now in place and visible at https://exercism.io/my/tracks/pharo

PR #36, opened 2018-08-02
Documentation refactoring

macta:
Had some time on the train with an iPad so focused on readability and trying to streamline the docs a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,121 +2,118 @@
 
 ![build status](https://travis-ci.org/exercism/pharo.svg?branch=master)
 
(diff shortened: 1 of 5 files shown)


macta:
This would close #18

samWson:
Looks good. I don't know if this is the place for it, but it might be a good idea to mention the Pharo Discord chat somewhere as a good place to get help.

macta:
I’ve merged this so it’s clearer how to load the dev baseline as @samwson has noticed issues trying to load it which we need to understand

PR #37, opened 2018-08-02
Fix the  generator to handle package tags vs. packages

macta:
Remember the last generated path (so regeneration is less painful)
Use RPackage organizer as per code review


macta:
Merging so we can do some proper testing as @samwson has been having issues

PR #41, opened 2018-08-04
Getting started instructions

macta:
Consolidate changes around #39 with an improved template and a code generator that leverages the configlet to pull in exercise instructions and pull in the exercise class comment as a hint.

NOTE: this diverges slightly from what we were initially doing where we copied the exercise instructions from exercism into the test class comment manually. I think the risk of duplication is too great.


dev/src/ExercismTools/AbstractNautilusUI.extension.st
```diff
@@ -2,21 +2,31 @@ Extension { #name : #AbstractNautilusUI }
 
 { #category : #'*ExercismTools' }
 AbstractNautilusUI class >> packagesExercismMenu: aBuilder [
-	<contextMenu>
 	<nautilusGlobalPackageMenu>
 	| package target |
 	target := aBuilder model.
 	(package := target selectedPackage) ifNil: [ ^ target ].
-	(package name beginsWith: 'Exercism') ifFalse: [ ^ target ].
+	((package name beginsWith: 'Exercism') or: [ package package name beginsWith: 'Exercism' ]) ifFalse: [ ^ target ].
```
macta: This should be refactored - its working around the fact that tags are different from packages and not true sub-packages

dev/src/ExercismTools/ExercismManager.class.st
```diff
@@ -30,8 +30,7 @@ ExercismManager class >> parseExercismFetchString: osOutputString [
 	
 	osOutputString regex: '\/\w*-?\w+$' matchesDo: [ :m |  ^m allButFirst ].
 	
-	"self error: 'Could not find exercism result'"
-	^'hello-world'
+	self error: 'Could not find exercism result in: ', osOutputString
```
macta: This is waiting for the exercism client to work on our track - at the moment we can't do anything

dev/src/ExercismTools/ExercismManager.class.st
```diff
@@ -53,6 +52,17 @@ ExercismManager class >> submitToExercism: packageOrTag [
 	UIManager default inform: cmd, ' -> ',result output, '.'
 ]
 
+{ #category : #exercism }
+ExercismManager class >> viewOnExercism: packageOrTag [ 
+
+	|  cmd result |
+		
+	cmd := 'open https://exercism.io/my/tracks/pharo'.
```
macta: In future - we might associate the exercise uuid with a test so we can jump to it directly - for now this opens a generic page.

Note: will need modification for windows - I think the command is run vs open?

dev/src/ExercismTools/ExercismManager.class.st
```diff
@@ -53,6 +52,17 @@ ExercismManager class >> submitToExercism: packageOrTag [
...
+				
+	result := PipeableOSProcess waitForCommand: cmd.
```
macta: A future story should do something with the results of process

macta:
@samWson not sure if this is just too much to review, but I'm going to merge it and accept any feedback ongoing (its all pretty much work in progress until we can try it end to end) and currently the exercism client is not letting us do it.

PR #47, opened 2018-08-04
WIP: Reverse string exercise

macta:
Initial attempt at the reverse string exercise - this uses recursion but I wonder if its too complicated or possibly there should be two potential solutions?


macta:
Needs to come after the Getting Started Instructions pr

dev/src/Exercism/ReverseString.class.st
```diff
@@ -0,0 +1,29 @@
+"
+Thist solution shows a linear approach to solving the problem, but also has an example of a recursive version.
```
samWson: Typo here: `Thist`.

dev/src/Exercism/ReverseStringTest.class.st
```diff
@@ -0,0 +1,58 @@
+"
+While there is a #reversed method for Strings, can you figure out how to do this with lower level iteration or streams? 
+
+Pharo has a neat code finder tool, press Shift-Enter  to activate it.
```
samWson: I like this suggestion here. Also another neat tool that I like is the Finders `Examples` tool. You give it an expected input and output, and it will find the method that does it if one exists. I found it's a bit tricky to use though and there is not much documentation on it, but it would be great to introduce people to it.

dev/src/Exercism/TwoFer.class.st
```diff
@@ -3,9 +3,9 @@ This is a sample solution for Exercism students.
 
 Of interest, in this example is that classes can define instance methods with the same name, but taking different numbers of parameters.
 
-A method which has no parameters, is called a unary method, and a method taking multiple parameters, each deliniated by a 'word ending with a ':', is called a keyword method.
+A method which has no parameters, is called a unary method, and a method taking multiple parameters, each deliniated by a word ending with a ':', is called a keyword method.
 
-There is a thrid type of method, which takes only 1 parameter but uses  a symbol (s) for a name (typically a mathematical one like: +, -, & etc) which is called a binary method.
+There is also a third type of method, binary, which takes only 1 parameter but uses a symbol(s) for a name (typically a mathematical one like: +, -, & etc).
 "
```
samWson: I like that your finding a way to introduce key Smalltalk rules on message types here. Would be great to include the precedence of messages too if you can squeeze it here or later in another exercise.

dev/src/Exercism/ReverseString.class.st
macta: good catch

dev/src/Exercism/ReverseStringTest.class.st
macta: Yes - that Finder tool is neat - you quickly outgrow the examples but its a real eye opener. Lets find a way to show that in a future exercise so we don't overwhelm people. Particularly as that and the test search are the only redeeming features of that tool (its very old - and probably should now be plugins for Spotter)

PR #54, opened 2018-08-06
Backup the exercism meta info when submitting

macta:
As tonel writer blows away the directory - we need to preserve the solution meta data


macta:
@samWson this final change should give us an end to end get an exercise and then submit workflow. It needs testing but my exercism account is now in a stuck state

PR #68, opened 2018-08-09
Fix environment issues in Pharo 6.1 that user may encounter

macta:
Fix refactoring in the debugger
Fix missing deny:equals for testing


macta:
Not essential stuff - but they were things I hit when following the philosophy we mention on the website. So I might propose putting them in (they are safe)

PR #70, opened 2018-08-09
Initial exercises - categorisation consistency improvements, small tweaks to readme text

macta:
When creating Leap - I spotted that we hadn't categorised consistently, so I think this is better example - tests in #testing, the sample class in #exercise.

Give it a review and see if you guys agree

### DIFF
--- a/dev/src/Exercism/HelloWorldTest.class.st
+++ b/dev/src/Exercism/HelloWorldTest.class.st
@@ -1,13 +1,17 @@
 "
-In Smalltalk you can run code (including tests) with broken methods and
-missing classes. When the VM encounters broken code, it raises a debugger
-which lets you explore the problem a little better.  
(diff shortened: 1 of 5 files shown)


dev/src/Exercism/HelloWorldTest.class.st
```diff
@@ -1,21 +1,25 @@
 "
-In Smalltalk you can run code (including tests) with broken methods and
-missing classes. When the VM encounters broken code, it raises a debugger
-which lets you explore the problem a little better.  
+If you are editting this file in a text editor and not in Pharo - 
+please review the [Getting Started](https://exercism.io/tracks/pharo/installation) section.
```
macta: Added this, as it dawned on me that Serge might have just edited the text (otherwise I'm not sure how he got the code into the image without the exercism menu?) Anyway - it seemed like a good reminder place for a newbie.

PR #80, opened 2018-08-12
Sum of Multiples exercise

macta:
First pass at sum of multiples. I followed other languages with a test for each case, but I do wonder if an array of test data might have been better.


macta:
Cool - we can do a second pass on all of them when we have a decent batch.

Should we make a checklist of them and discuss merits and who has taken one to do next?

samWson:
> Should we make a checklist of them and discuss merits and who has taken one to do next?

@macta of course. I think we may already have an issue open for something like this? If not we can just make a new one.

Issue #72, opened 2018-08-10
pharo maintainers link shows 0 - decide who to display


macta:
We need to fill out the maintainers.json file and set show_on_website to true, for those maintaining.

samWson:
I can pick this one up. I'm a bit low on time and this is a quick win. I've had a look at the Ruby one for reference.

@macta I see your already in there. Do you want any additional info e.g. bio, visible on website?

@bencoman I take it you will want to be listed there too? Have a look at the [file](https://github.com/exercism/pharo/blob/master/config/maintainers.json) for the kind of information that can go there. Also let me know if you want to be visible on the website.

If I've left anyone else out who wants to be a maintainer, please leave a comment below.

bencoman:
Added my details.  I don't feel I've done enough lately or for the near future to take credit being visible on the web site.  I'd love to do more and hope to have time in a few weeks.

macta:
I’ll add something like you - but I’m in the road at the moment so not been able to easily update mine yet.

At least we are showing 1 maintainer which is a good start.

Looks nice on the site too!

macta:
I’ve done my bio too. Closing 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #84, opened 2018-08-14
Add bencoman entry

bencoman:
Issue #72


macta:
You've upped the anti with a blog link...

PR #85, opened 2018-08-14
Acronym exercise

macta:
add's another exercise (interestingly - the bulk of this one was written on an iPad - just hand editting tonel as I didn't have a laptop handy)

There are some doc changes I noticed when I was browsing


macta:
This felt like a natural follow on from reverse-string (although the other tracks seemed to follow on from two-fer which feels very weird to me) - in my mind its a similar kind of problem with a bit more complexity?

samWson:
@macta Looks good. I'm getting more used to Iceberg so I pulled this branch and ran the tests on my image. All green. Merge when your happy.

PR #92, opened 2018-08-29
Integrate Ben's changes (with minor modifications) to retrieve source…

macta:
… files wihout OSProcess

This is a quick port of what Ben suggested. Its a bit rough and ready, but it would get us a bit further.

NOTE: we don't have to use NeoJson as STON is backwards compatible for such a small bit of Json

### DIFF
--- a/dev/src/ExercismTools/ExercismManager.class.st
+++ b/dev/src/ExercismTools/ExercismManager.class.st
@@ -28,28 +28,26 @@ ExercismManager class >> welcome [
 
 { #category : #exercism }
 ExercismManager >> fetchFromExercismTo: package [
-	| result cmd exercise tonelLocator reader pkgName root exercisePkg |
+	| result exercise pkgName root exercisePkg |
 	
-	(exercise := UIManager default request: 'Enter a valid exercism exercise (e.g. hello-world):') ifNil: [ ^ self ].
-	cmd := 'exercism download --track=pharo --exercise=' , exercise.
+	self verifyToken ifFalse: [^self].
 	
+	(exercise := UIManager default
+		request: 'Enter a valid exercism exercise (e.g. hello-world):')
+		ifNil: [ ^ self ].
+		
 	UIManager default inform: 'Loading: ' , exercise.
-	
-	result := PipeableOSProcess waitForCommand: cmd.
-	self verifyExercismResult: result.
-	
-	tonelLocator := FileLocator cwd / exercise.
-	(reader := TonelReader on: tonelLocator fileName: tonelLocator pathString) loadDefinitions.
-	reader definitions
-		do: [ :def | 
-			def load.
-			def isOrganizationDefinition
-				ifTrue: [ pkgName := def categories first ] ].
-			
+	(result := ExercismDownloader exercise: exercise)
+		detect: [ :def | def isClassDefinition ]
+		ifFound: [ :def | pkgName := def category ].
+		
 	root := RPackageOrganizer default packageNamed: 'Exercism'.
-	exercisePkg := root packages detect: [ :p | p categoryName = pkgName ].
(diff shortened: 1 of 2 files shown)


macta:
@bencoman not sure why you don't appear as a potential reviewer

dev/src/ExercismTools/ExercismManager.class.st
```diff
@@ -93,14 +91,14 @@ ExercismManager >> submitToExercism: packageOrTag [
 ]
 
 { #category : #helper }
-ExercismManager >> verifyExercismResult: osProcessResult [
-
-	osProcessResult succeeded
-		ifTrue: [ (osProcessResult output beginsWith: FileLocator cwd pathString)
-				ifFalse: [ self error: 'Pharo must be run from the default exercism directory!' ].
```
samWson: If your removing this error, does that mean you can run Pharo from outside the exercism directory and still have the whole thing work?

samWson:
> @bencoman not sure why you don't appear as a potential reviewer

I just checked to be sure, @bencoman you are listed as a contributor so I'm not sure why you don't appear as a potential reviewer.

@macta fast work as usual. Still I enjoy getting to read and understand your work.

dev/src/ExercismTools/ExercismManager.class.st
macta: Yes the implication (when we get the submission part of this change done) would be that we aren't tied to the exercism path (which would be a nice side effect and make installation instructions much easier for windows. For osx, I would still be tempted to zero conf install pharo in that directory anyway as it makes us look much more like normal languages - but if you wanted to follow the windows instructions and use pharolauncher that would just work too).

We may hit a future point where we need to save some data files that go along with an exercise somewhere - and so I anticipate some future work in this area unless we have a different way of storing data - e.g. like Seaside does by putting it encoded into a bytearray. But lets not overthink it as this point.

macta: In the end, the bit you guys figured out - slotted in quite easily

bencoman:
On Wed, 29 Aug 2018 at 16:35, Samuel <notifications@github.com> wrote:

> @bencoman <https://github.com/bencoman> not sure why you don't appear as
> a potential reviewer
>
> I just checked to be sure, @bencoman <https://github.com/bencoman> you
> are listed as a contributor so I'm not sure why you don't appear as a
> potential reviewer.
>
Last week I PR'd a simple doc change and thought I would be able to \
Accept it myself,
but seemed to not have permission.  Is there some constraint you can't
accept your own PRs?
Otherwise this seems related to not appearing as a potential reviewer.

macta:
@bencoman yeah you should be able to approve your own pr - Sam and I can. I normally try to give a few days for one of you to take a look and then I approve it myself as i know we are all a bit busy and trying to move this project forward.

Something must be set incorrectly

dev/src/ExercismTools/ExercismManager.class.st
samWson: >  For osx, I would still be tempted to zero conf install pharo in that directory anyway as it makes us look much more like normal languages

I really appreciate that simple `curl` commands and a zero conf install will at least put everyone on some common ground. It's a bit frustrating that I couldn't get `curl` to make it through the Window Subsystem for Linux. I'm apprehensive about putting potential beginners through too many hoops installing tools just to get a working environment, so I just had a quick look to see if there are any native Windows tools like curl or wget that we can expect to be available by default. It turns out there is something called [BITSAdmin Tool](https://docs.microsoft.com/en-us/windows/desktop/Bits/bitsadmin-tool). It could be an option if we want to try a zero conf install on Windows later on.

Great work by the way. Getting closer to the goal.

PR #98, opened 2018-09-06
Move setup to class initialization. Create Welcome class programatically.

bencoman:
Discussion in exercism/pharo#97


bencoman:
For initial testing, if you load this PR into an image that already has the ExercismManager class, I believe the class initialization won't trigger.  You'll need to manually run ```ExercismManager initialize```.
Final testing should be in a fresh image and it should run automatically with...
```Metacello new 
 baseline: 'Exercism'; 
 repository: 'github://exercism/pharo:master/dev/src';
 load.```

**Comment by bencoman** [comment]
Here is a recipe to test the class initialization loading this PR (actually its source branch) into a fresh image.  In Pharo 6.1...
Downloaded http://files.pharo.org/platform/Pharo6.1-win.zip
Updated Iceberg per https://github.com/pharo-vcs/iceberg#update-iceberg
Iceberg 
  > Settings > Use custom keys = Enabled
  > Add > Clone remote repository
      Remote URL = git@github.com:exercism/pharo.git
  ==> "pharo" repo appears, status "Not loaded"
  > "pharo" repo > Repository > Add remote 
         > Remote-Name=ben; Remote-URL=https://github.com/bencoman/pharo-exercism.git
  ==> "ben" appears under "Remotes"
    > "ben" > "welcome-setup" > Checkout branch > Checkout
    ==> Branches > "welcome-setup" appears, 
             and back in main Repositories window Branch=welcome-setup
  > "pharo"(welcome-setup) > Packages > "ExercismTools" > Load
  ==> TaDah! Class "Welcome" has been created under package "Exercism" 
      and "ExercismTools" package is clean.

**Comment by bencoman** [comment]
IIUC the development environment is not loaded. In a completely fresh Pharo 61 image, doing only this....
```
Metacello new 
 baseline: 'Exercism'; 
 repository: 'github://bencoman/pharo-exercism:welcome-setup/dev/src';
 load.
```
ends up looking like this...

a06-welcome-setup

which I presume from the INSTALLATION instructions replicates how a student loads Exercism.

macta:
It’s not quite the same - the command line version (which I expected windows to also support - but sadly isn’t quite there with zero conf ) opens a system browser with Welcome selected (although not the comment pane with details, selected).

This is why I hadn’t used class initialize (it seemed a bit risky), and also I wasn’t expecting the dev mode (when creating exercises) to do this and pollute proper exercises with Welcome.

However - a lot changed as I developed it, and we ended up with exercises all in an Exercism package which we then export to exercism git. Hence me now wondering if we might be able to do something less clever and overlay packages somehow (but it gets tricky - as we’re not trying to teach people iceberg or metacello , we just expect them to be familiar with exercism). In effect we are sort of automating the first load of an exercise to demonstrate how it looks and also give you something to right click on for the exercism menu.

I’ll try and get a photo for you to show it now on OS X - as it might not have been so obvious for you.

macta:
The startup looks like this on OSX (well close, as this image has a solution defined and I had also loaded another example - but it gives the idea.

image

bencoman:
How about doing the welcome with the Help system. e.g. "WelcomeHelp open"
When the student tools are loaded we could pop up our custom WelcomeExercism page.

bencoman:
Reference List for discussing options to bootstrap students to use ExercismTools 
(edit this entry to append any more you think of).

1. Evaluate Playground code + manual initialize ```Metacello .... load.  ExercismManager welcome```
1. Evaluate Playground code + auto initialize ```Metacello .... load.```
1. Use Pharo Catalog + auto-initialize ```World menu > Tools > Catalog Browser > "Exercism" > Install stable version```
1. Link to our own "Exercism Image" built by CI

bencoman:
Regarding creation of the "Exercism" package to provide the Exercism-Menu.  I don't see any risk in doing this in the class-side initialize. To deal with the package being deleted by a student, it could be reinstated from the instance-side initialize of ExercismCommand (currently this is done at the top of ExercismDownload>>installExercise).

Also, maybe we could add a trimmed down Exercism-Menu to the ExercismTools package to just do "Fetch".

bencoman:
Considering complete newbies, the priorities are: QuickStart, Simplicity, Reliability.  My evaluation...
[2.] Catalog could be nice, but adds a dependency on SmalltalkHub which has varying reliability.
[3.] Own Image might provide the most reliable quick-start, but more effort for us to maintain.  
[4.] PharoLauncher adds a level of complexity and dependency ==> less quick-start.  
[5.] Help Integrated would be very nice, but long term 
[1.] Playground code minimizes dependencies ==> reliable. Also immediately shows how evaluated code can modify the environment.

So for the moment I prefer [1.], maybe [3.] later.  Along those lines, for the first exercise only (i.e. for HelloWorld) I'd like our walk through instructions to show the Download/Submit from Playground (and thereafter from the GUI) - just to showcase this aspect of Pharo's uniqueness.

macta:
#1 is effectively what the Pharo track docs say - and is my preferred suggestion. However - to be more like other languages, I wrote our docs to do this from the command line (which i think is better). 

Even with all the windows troubles - you can still run command line instructions for windows (even if it doesn’t do zero conf). Can we just update the instructions for windows about how to download Pharo and then give the same command line instructions?

It makes us look just like other environments and you also don’t have explain how to execute playground code etc when people are just getting started.

We should introduce those concepts through the exercises. That is the exercism way (in my mind). We educated by discovery,

bencoman:
>
> Reviewing this one - I can't see how this would work if you have a user
> (default) baseline and dev baseline for the use of #initialize? As
> initialize will always run, and you don't want it to for the latter case.
>
> I think programmatically creating welcome for user's is fine (which has
> the side effect of creating the Exercism package that users can then load
> exercises into). So I like that suggestion (vs. the slightly hacky move we
> do now.
>
> However in the dev baseline - you don't want to end up committing Welcome
>
Another reason why I wanted split repos:
* dev/execrises
* user/tools

> (this is why the command line instructions invoke #welcome as part of the
> loading instructions - which aren't the same for dev mode). Rather than try
> to detect what baseline is being loaded and being too magical, why don't we
> keep it that way - and as per my previous comments about loading on Windows
> - we just get all users to load the initial baseline from the command line
> the same way (v.s. linux/osx users do command line and windows users do
> something else).
>
> My take is that the Exercism community love the command line - and we can
> embrace it too, until users understand that Pharo is a kind of fancy repl
> itself once they dig in.
>
We differ views slightly here.  Mine is that the Playground provides a the
"command-line" equivalent experience people want.  My aim with having the
ExercismCommand hierarchy is to demonstrate early that Pharo IDE is "not
just an IDE" -- its "different" so get curious.

But anyway, my thinking has moved on from this.  I think we don't need the
Welcome class at all, since...
* the welcome info will be better presented in a Help file - I'll submit an
example PR for discussion.
* safer for every ExercismDownload to create the Exercism package if it
doesn't exist, as currently happens
* instructions for first exercise download "Hello-World" can be done
through Playground to demonstrate early on our command-line-like facility
* the ExercismTools package can provide the identical menu as the Exercism
package before that is created

...so I will close this issue.

Issue #97, opened 2018-09-06
Discussion - initial impressions


bencoman:
Just installed the Pharo client for the first time and thought it would be good to record my initial impressions. 

I did this...
1. Downloaded Pharo Windows Standalone... http://files.pharo.org/platform/Pharo6.1-win.zip
2. Updated Iceberg... https://github.com/pharo-vcs/iceberg#update-iceberg
3. Installed Exercism... https://github.com/exercism/pharo/blob/master/docs/INSTALLATION.md
except skipped running "ExercismManager welcome"

System Browser shows...
![a01-post-load](https://user-images.githubusercontent.com/1713447/45140570-c9703000-b1e5-11e8-9717-d72eb85dd0ff.png)

and after now evaluating ```ExercismManager welcome```, System Browser shows...
![a02-post-welcome](https://user-images.githubusercontent.com/1713447/45162441-9ac37b00-b220-11e8-9f75-679ece9bf1b7.png)

Great idea, but the downside of relocating a class out of the ExercismTools package is:
* ExercismTools package becomes dirty
* running it a second time (because someone will try - at least I'm looking at myself) shows an error message "package already exists" and opens a debugger "#moveClass:toTag: was sent to nil"

Instead of delivering a Welcome class in ExercismTools, we can create it programatically.
Also, we don't need the user to explicitly run ``````ExercismManager welcome```.  
It can be done by the class initialization when ExercismManager class is first loaded.

bencoman:
Really cool idea to add an "Exercism" menu, but I had trouble finding it near the bottom of the list. I think it works better at top of the menu.
![a04-exercism-menu](https://user-images.githubusercontent.com/1713447/45196948-57552500-b291-11e8-9465-679270573351.png)

bencoman:
I like the way that fetching the first exercise asks for the student token,
but I felt lazy** and didn't want manually to browse to the Exercism settings page to get it.
So I spent a few hours working out how to provide a clickable link - something new to me but was optimistic would work.

** one of the programming virtues, but [was it worth the time?](https://xkcd.com/1205/)

bencoman:
I'll open another issue for subsequent ideas.

PR #99, opened 2018-09-07
Move "Exercism" package menu to top.

bencoman:
I had trouble seeing the menu item when was near the bottom.


bencoman:
Discussion #97

PR #100, opened 2018-09-07
Clickable link to token

bencoman:
In the dialog box requesting the user enter their Exercism student token, provide a clickable link to get this from the Exercism Settings web page.

Testing from playground... ```ExercismDownloader configureToken```

PR #101, opened 2018-09-08
Ready multiple commands

bencoman:
Note that this is downstream from PR #100 "Clickable link to token" 
(MorphicUIManager StaticTextMorph TextEntryDialogWindow UITheme)

I'm not sure why the menu location change (AbstractNautilusUI.) is showing up as an addition here, 
since PR #99 was integrated into master, and I've double-checked the same code is in both branches,
and in my pharo git repo from the command-line  `git diff c334bf8 ebd1549` shows no differences.

Renamed ExercismDownloader to ExercismDownload since that is more "command-like".
i.e. the CLI command is not `exercism downloader --track=xxxx --exercise=yyyy`
(Sorry I should have done that as a separate commit)

Adds ExercismCommand as a superclass to ExercismDownload and refactored functionality common to ExercismSubmit (i.e. the API key and http client)

Some code from  ExercismManager>>fetchFromExercismTo:  belonged better in ExercismDownload.
and from ExercismManager>>verifyToken  in ExercismCommand.


bencoman:
Yes, I realize I'm bending the rules, noted in [this comment](https://github.com/exercism/pharo/pull/101/files#diff-8b0cdc2f98af299a90c80cf87795dc61R22).
What I'd like is something that reads well as a command from Playground, in a form as close as possible to the CLI.   These commands could be written as links in help file documentation (not sure yet how useful that would be) and I have half a plan later to write a blog "How to build an Exercism client" and a step along the way would be running commands from Playground before doing any GUI stuff.

I'd like to refactor some temporary variables from ExercismDownload>>track:exercise:
to be instance variables,  so its easier to see a self contained download - for both debugging and demonstration.  The current temporary variables are too transient and you can't see the action without adding a breakpoint.

macta:
Lets run with it - there's not a lot of us, and we can always rename it back if it does't work out.  I come from a time of noun's and verbs OO - and so an XXXDownload doesn't read so well to me  (possibly a suffix like  Command or Operation helps). Anyway - lets not get diverted - the real task is to get it working again on OSX (and actually working reliably on Windows).

As a small point, I think delaying renaming while changing behaviour makes it easier to diff changes going forward.

dev/src/ExercismTools/ExercismCommand.class.st
```diff
@@ -0,0 +1,96 @@
...
+   ""#queryAt: 's accumulate,  and are
+    cleared by #path:  when reusing connection""
+   
+Internal Representation and Key Implementation Points:
+    ApiPath - path string required by server e.g. '/v1/solutions/latest'
+    ApiToken - string e.g. 'd4a3xxxx-secret-xxxxxxxx9c56'
+
+One concern is the key being permanently retained in Image, in a corner case where the  Image may be  shared with third-parties.
+Perhaps later we should consider storing key in voltatile memory and/or external file.  btw, where does the CLI store its key?
+"
```
macta: Yes - in image caching was a quick stab at this. V2 I would say - for now bigger fish to fry

dev/src/ExercismTools/ExercismCommand.class.st
```diff
@@ -0,0 +1,96 @@
...
+
+{ #category : #gui }
+ExercismCommand class >> configureToken [
+   |requestText linkText token url |
+	url := 'https://exercism.io/my/settings'.
+   linkText := url asText
+		addAttribute: (TextAction new actOnClickBlock: [ExternalWebBrowser new open: url]);
+		addAttribute: TextColor blue;
+		addAttribute:TextEmphasis underlined.
+   requestText := 'CLI token available at  ' asText, linkText.
```
macta: I like this, its neat.

dev/src/ExercismTools/ExercismCommand.class.st
```diff
@@ -0,0 +1,96 @@
...
+{ #category : #internal }
+ExercismCommand class >> newHttpClient [
+    ^ ZnClient new 
+          http;
+          host: 'api.exercism.io';
+          headerAt: 'Authorization' put: 'Bearer ' , ApiToken
+]
+
+{ #category : #internal }
+ExercismCommand class >> path [
```
bencoman: Thx for the approve @macta. It is easier to move forward with this given PR and I'm open to discussing reverting the name later if it doesn't settle in well for you.  Now I discovered this #path method here should be on the instance side - which highlighted the lack of a test.  So I'm in the middle of adding one, which  I'll commit it to this PR for that before merging.  This needed some refactoring to facilitate testing, but I'll push that through without waiting.  Concerns can be addressed in another PR.

PR #102, opened 2018-09-09
Fix ExercismDownload for real.

bencoman:
After refactoring to facilitate a unit test for instance side, I missed fixing class side to match.
btw, also now doesn't require #welcome to be run.  Creates "Exercism" package on demand.
https://github.com/exercism/pharo/compare/fix-ExercismDownload-for-real?expand=1#diff-8b0cdc2f98af299a90c80cf87795dc61R90


macta:
Thanks for starting some tests - I’ve felt very guilty about not having more of them

Issue #96, opened 2018-09-06
Submitting solutions direct from Pharo


bencoman:
Following on from success with [downloading exercises directly into Pharo IDE](https://github.com/exercism/pharo/issues/32), now need to work on submitting solutions.

bencoman:
C:> exercism --verbose submit hello_world.py
========================= BEGIN DumpRequest =========================
PATCH /v1/solutions/53cdc7dc803248d19da68eb79827c5d1 HTTP/1.1
Host: api.exercism.io
Authorization: Bearer d4a3xxxx-secret-xxxxxxx59c56
Content-Type: multipart/form-data; boundary=350c70047f7127f7cdc3a6a3bc43eb4b6ffcbfe0bd5ac42b0c04f64b2a2b
User-Agent: github.com/exercism/cli v3.0.8 (windows/amd64)

--350c70047f7127f7cdc3a6a3bc43eb4b6ffcbfe0bd5ac42b0c04f64b2a2b
Content-Disposition: form-data; name="files[]"; filename="hello_world.py"
Content-Type: application/octet-stream

def hello():
    return 'Hello, World!'

--350c70047f7127f7cdc3a6a3bc43eb4b6ffcbfe0bd5ac42b0c04f64b2a2b--

========================= END DumpRequest =========================


========================= BEGIN DumpResponse =========================
HTTP/1.1 201 Created
Transfer-Encoding: chunked
Cache-Control: max-age=0, private, must-revalidate
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Thu, 06 Sep 2018 01:43:24 GMT
Etag: W/"44136fa355b3678a1146ad16f7e8649e"
Referrer-Policy: strict-origin-when-cross-origin
Server: nginx/1.10.3 (Ubuntu)
Set-Cookie: _exercism_session=bv7WKRITuNVfpV05eVvKRolymVq2TrSg9Frz9VMzUTEhnnZURyaQXHxI1HgGfEgbtcO1X26NLTSibYCCS%2BegOKIIEQT8BgPAGpj3JNQpQ%2BLpPPIcrJh1vy0K181suVWL8OR2Y5AbNuK6e3%2FabWC9CMmoWQ%3D%3D--4qY4Frvj7uilbQuN--7rnOmHrU815uFvM7%2B4xNFQ%3D%3D; domain=.exercism.io; path=/; HttpOnly
X-Content-Type-Options: nosniff
X-Download-Options: noopen
[...]

PR #104, opened 2018-09-16
Store metadata in package "Properties" class var.

bencoman:
We need somewhere to store the exercise metadata to be used by Submit command.  Some options are: 
- as a class comment
- as a special method
- a Manifest-like class
- in the package's "Properties" class variable

I'd first like to try the last one since it has the least visual distraction for students.
To test this PR...
1. Create an empty "Exercism" package
2. Right-click on "Exercism" package > Exercism > Fetch Exercise = hello-world
3. Right click on "HelloWorld" tag> Exercism > View metadata


bencoman:
This might not be the best way to store solution/exercise metadata (e.g. a custom RPackageTag subclass may be better) but I'm merging since it is sufficient to provide "Solution-ID" required by Sam to implement ExercismSubmit for issue #96.

PR #106, opened 2018-09-17
Allow generated ZnMimeParts to be accepted by Exercism

samWson:
This one is preparation for finally fixing #96. @bencoman this manual patch is as you showed in the same issue. I just manually added the missing white space and adopted the method by changing its protocol to `*ExercismTools`. I'm guessing when our code is loaded into a fresh image, this change will get merged and patch the method for us?

### DIFF
--- a/dev/src/ExercismTools/ZnMimePart.extension.st
+++ b/dev/src/ExercismTools/ZnMimePart.extension.st
@@ -1,0 +1,14 @@
+Extension { #name : #ZnMimePart }
+
+{ #category : #'*ExercismTools' }
+ZnMimePart class >> fieldName: fieldName fileName: fileName entity: entity [
+	"Pathnames are often silenty encoded using UTF-8,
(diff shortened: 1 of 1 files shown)


bencoman:
This is exactly what I was thinking of, but in hindsight adopting that method dirties the Zinc-HTTP package for no good reason.  Its not like this method is buried deep in a call stack. We use it directly, so can just copy the existing method to ```ZnMimePart class >> exercismFieldName:fileName:entity:``` and use that.  I think thats what Tim indicated [here](https://github.com/exercism/pharo/issues/96#issuecomment-421789420) but I wasn't paying attention.  Then we don't dirty the Zinc package.  

Sam, Rather than ask you to resubmit due to my error, I've created [PR#107](https://github.com/exercism/pharo/pull/107). Could you review/integrate?

macta:
Yes - just a prefixed extension method will work fine (that’s what I was trying to say)

@bencoman did you report this to Sven yet? I think the spec is unclear and it’s something you have to deal with

PR #110, opened 2018-09-26
Issue 96: Submit exercises via HTTP patch method

samWson:
I broadly understand what this thing needs to do now. I won't be able to spend anytime on this tomorrow so I'm placing this pull request here to make sure I'm on the right track. Hopefully I'll be able to get back to it this weekend.

This is definitely a work in progress. A few more helper methods will probably be made, plus getting rid of some hard coded values. 

I used the Acronym exercise as an example, which means I could not test submitting because I have not downloaded the exercise metadata, which I can't do because I haven't unlocked the exercise ¯\_(ツ)_/¯ .

I've left a few comments in the source code with my thoughts and a few problems I haven't yet solved. Let me know what you think.

I'm also irritates me that I don't have any tests of the public interface. For lack of a good mocking framework it will take a bit of work to mock Exercisms API, and the OS file system.

### DIFF
--- a/dev/src/ExercismTools/ExercismCommand.class.st
+++ b/dev/src/ExercismTools/ExercismCommand.class.st
@@ -71,7 +71,7 @@ ExercismCommand class >> configureToken [
 ExercismCommand class >> configureToken: your_CLI_token [
     "Get your_CLI_token at https://exercism.io/my/settings"
     ApiToken := your_CLI_token.
-    ApiPath := '/v1/solutions/latest'.
+    ApiPath := '/v1/solutions'.
 ]
 
 { #category : #testing }
@@ -85,6 +85,13 @@ ExercismCommand class >> newHttpClient [
           https;
           host: 'api.exercism.io';
           headerAt: 'Authorization' put: 'Bearer ' , self token
+]
+
+{ #category : #command }
+ExercismCommand class >> reset [
+	ApiPath := nil.
+	ApiToken := nil.
+	
 ]
 
 { #category : #'as yet unclassified' }
@@ -106,4 +113,9 @@ ExercismCommand >> initialize [
 { #category : #internal }
 ExercismCommand >> path [
 	^ApiPath copy
+]
+
+{ #category : #'test support' }
+ExercismCommand >> responseCode [
+	^ httpclient response code
 ]
--- a/dev/src/ExercismTools/ExercismDownload.class.st
+++ b/dev/src/ExercismTools/ExercismDownload.class.st
@@ -38,7 +38,7 @@ ExercismDownload class >> exercise: exerciseId [
 	^(self new track: 'pharo' exercise: exerciseId) execute
 ]
 
-{ #category : #'instance-creation' }
+{ #category : #command }
 ExercismDownload class >> track: trackId exercise: exerciseId [
 	^(self new track: trackId exercise: exerciseId) execute
 ]
@@ -71,7 +71,7 @@ ExercismDownload >> filesContentMap [
 ExercismDownload >> getSolutionData [
 	| responseString response |
 	httpclient 
-		path: self path;
+		path: self path, '/latest';
 		queryAt: 'track_id' put: track;
 		queryAt: 'exercise_id' put: exercise.
 		
@@ -81,6 +81,7 @@ ExercismDownload >> getSolutionData [
 	response at: 'error' ifPresent: [ :error | self error: (error at: 'message') ].
 	
 	solution := response at: 'solution'.
+	ExercismManager solutionData at: exercise put: solution.
 ]
 
 { #category : #internal }
@@ -94,16 +95,7 @@ ExercismDownload >> installExercise [
 	[	:filename :contents | (filename endsWith: '.st') ifTrue: 
 		[	|parser|
 			parser := TonelParser on: contents readStream.
-         		parser document do: 
-			[	:def | 
-				def load.
-				def isClassDefinition ifTrue: 
-				[	|tag|	
-					"Store solution metadata."
-					tag := exercismPackage classTagNamed:  (exercismPackage toTagName:  def category).
-					exercismPackage ensureProperties at: tag put: solution.
-				].
-			].
+         	parser document do: [:def | def load].
 		].
 	].
 	SystemAnnouncer uniqueInstance announce: (RPackageRegistered to: exercismPackage).
@@ -117,5 +109,5 @@ ExercismDownload >> solution [
 { #category : #accessing }
 ExercismDownload >> track: aStringTrack exercise: aStringExercise [
 	track := aStringTrack.
-	exercise := aStringExercise.
+	exercise := aStringExercise asKebabCase.
 ]
--- a/dev/src/ExercismTools/ExercismManager.class.st
+++ b/dev/src/ExercismTools/ExercismManager.class.st
@@ -72,9 +72,7 @@ ExercismManager >> submitToExercism: packageOrTag [
 				collect: [ :cls | 
 					self pathStringFor: (submissionDirectoryRef / (writer fileNameFor: cls asClassDefinition)) ]).
 				
-	cmd := 'exercism submit ' , filesToSubmit.
-	result := PipeableOSProcess waitForCommand: cmd.
-	result succeeded ifFalse: [ self error: 'Unable to submit solution, CLI reported ', result outputAndError printString ].
+	ExercismSubmit exercise: exerciseName.
 	
 	UIManager default inform: 'Successfully submitted solution!'
 ]
--- a/dev/src/ExercismTools/ExercismManager.class.st
+++ b/dev/src/ExercismTools/ExercismManager.class.st
@@ -1,9 +1,12 @@
 "
-I am  responsible for managing the interaction with the lower level Execercism command line tool.
+I am  responsible for managing the interaction with the lower level Exercism command line tool.
 "
 Class {
 	#name : #ExercismManager,
 	#superclass : #Object,
+	#classVars : [
+		'SolutionData'
+	],
 	#category : #ExercismTools
 }
 
@@ -12,6 +15,11 @@ ExercismManager class >> default [
 	"answer a defulat manager"
 	
 	^self new
+]
+
+{ #category : #exercism }
+ExercismManager class >> solutionData [
+	^ SolutionData ifNil: [ SolutionData := Dictionary new ]
 ]
 
 { #category : #exercism }
--- a/dev/src/ExercismTools/ExercismSubmit.class.st
+++ b/dev/src/ExercismTools/ExercismSubmit.class.st
@@ -16,71 +16,93 @@ Class {
 	#name : #ExercismSubmit,
 	#superclass : #ExercismCommand,
 	#instVars : [
-		'exercise',
-		'exerciseMap'
+		'classes',
+		'exerciseMap',
+		'packageTag'
 	],
 	#category : #ExercismTools
 }
 
-{ #category : #accessing }
+{ #category : #command }
+ExercismSubmit class >> class: aClass [ 
+	^ self classes: { aClass }
+]
+
+{ #category : #command }
+ExercismSubmit class >> classes: classes [
+	^ (self new classes: classes) execute
+]
+
+{ #category : #command }
 ExercismSubmit class >> exercise: exerciseId [
-	^ (self new exercise: exerciseId) execute
-]
-
-{ #category : #accessing }
-ExercismSubmit class >> solutionDataForClass: aClass [ 
-	| exercismPackage classTag |
-	exercismPackage := RPackageOrganizer default 
-		packageNamed: 'Exercism' 
-		ifAbsent: [ self error: 'No exercises downloaded.' ].
-	classTag := exercismPackage classTagForClass: aClass.  
-	^ exercismPackage ensureProperties at: classTag.
-	
-]
-
-{ #category : #accessing }
-ExercismSubmit class >> solutionDataForExercise: exerciseString [ 
-	| exercismPackage |
-	exercismPackage := RPackageOrganizer default 
-		packageNamed: 'Exercism' 
-		ifAbsent: [ self error: 'No exercises downloaded.' ].
-	exercismPackage ensureProperties 
-		detect: [ :solutionData | ((solutionData at: 'exercise') at: 'id') = exerciseString ] 
-		ifFound: [ :solutionData | ^ solutionData ] 
-		ifNone: [ self error: 'Exercise ''' , exerciseString , ''' not downloaded.'].
-	
+	"Submit whole package-tag specified by exerciseId."
+	"By default, don't submit TestCases with solution (if needed use #classes: direct)"
+	|rootPackage packageTag solutionClasses|
+	rootPackage := RPackageOrganizer default packageNamed: 'Exercism'.
+	packageTag := rootPackage classTagNamed: exerciseId kebabAsCamelCase.
+	solutionClasses := packageTag classes select: [:class | (class new isKindOf: TestCase) not].
+	(solutionClasses size > 0) ifFalse: [  self error: 'No classes to submit' ].
+	^ self classes: solutionClasses
 ]
 
 { #category : #internal }
 ExercismSubmit >> buildRequest [
-	| solutionEntity solutionPart multiPartFormDataEntity solutionId |
-	"Assumption: There is only one entity to make which is from the class with the same name as the exercise."
-	solutionEntity := ZnByteArrayEntity bytes: (exerciseMap at: exercise capitalized asSymbol).
-	solutionPart := ZnMimePart exercismFieldName: 'files[]' fileName: exercise entity: solutionEntity.
-	multiPartFormDataEntity := ZnMultiPartFormDataEntity new addPart: solutionPart.
-	solutionId := (self class solutionDataForExercise: exercise capitalized) at: 'id'.
+	| solutionClasses multiPartFormDataEntity solutionId exercise |
+	multiPartFormDataEntity := ZnMultiPartFormDataEntity new.
+	classes do: [ :solutionClass | |solutionPart solutionEntity|
+		solutionEntity := ZnByteArrayEntity bytes: (exerciseMap at: solutionClass name asSymbol).
+		solutionPart := ZnMimePart exercismFieldName: 'files[]' fileName: solutionClass name, '.st' entity: solutionEntity.
+		multiPartFormDataEntity addPart: solutionPart.
+	].
+	
+	solutionId := (ExercismManager solutionData at: packageTag name asKebabCase) at: 'id'.
 	
 	httpclient 
 		url: ApiPath , '/' , solutionId;
 		entity: multiPartFormDataEntity
 ]
 
+{ #category : #accessing }
+ExercismSubmit >> classes: arrayOfClasses [
+	"Simplification - only submit classes for one exercise at a time"
+	| firstPackage rootPkg|
+	rootPkg := RPackageOrganizer default packageNamed: 'Exercism'.
+	firstPackage := nil.
+	arrayOfClasses do: [ :c | |package|
+		c isClass ifFalse: [ self error: c printString, ' is not a class.' ].
+		package := rootPkg classTagForClass: c.
+		firstPackage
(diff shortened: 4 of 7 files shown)


dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -51,3 +60,32 @@ ExercismSubmit class >> solutionDataForExercise: exerciseString [
 		ifNone: [ self error: 'Exercise ''' , exerciseString , ''' not downloaded.'].
 	
 ]
+
+{ #category : #internal }
+ExercismSubmit >> buildRequest [
+	| solutionEntity solutionPart multiPartFormDataEntity solutionId |
+	solutionEntity := ZnByteArrayEntity bytes: exerciseContents.
+	solutionPart := ZnMimePart exercismFieldName: 'files[]' fileName: 'Acronym.class.st' entity: solutionEntity.
```
samWson: Hard coded file name. There might be multiple files for a solution i.e. the class side and instance side of a class. I'm not that familiar with using Smalltalk on file systems. This is a problem I haven't solved yet.

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -51,3 +60,32 @@ ExercismSubmit class >> solutionDataForExercise: exerciseString [
...
+	httpclient patch.
+	
+]
+
+{ #category : #accessing }
+ExercismSubmit >> exercise: aStringExercise [
+	| path |
+	exercise := aStringExercise.
+
+	path := (Path from: (aStringExercise asKebabCase)) / '.pharo' / aStringExercise capitalized , 'class.st'.
```
samWson: Again solution hard coded here: `class.st`. Seems my number one problem is getting data from all the possible files a exercise could have. I figure for an exercise we will need to grab all files in the directory ending with `.st` and make a mime part for each, excluding the test class and the package file. I don't have much experience walking file systems so its a new problem to me. 

Also this code assumes that Pharo is being run from the `Exercism/pharo` directory.

macta:
I havent had a chance to look at your code (family duties are in the way),but from your comments I worry you may have gone astray... aren’t we now divorced from the file system? You should be able to iterate over the classes in an exercise sub-pkg/tag and generate the source for each and attach it as a “file” (no real file system should be needed).

Pissiblybim mossing something - but that’s what went through my head as a soln. (Just as retrieve doesn’t get any real file either).

There is also some metadata stuff we need to do as well - by again it should be file-less

bencoman:
>
> which means I could not test submitting because I have not downloaded the
> exercise metadata, which I can't do because I haven't unlocked the exercise
> ¯_(ツ)_/¯ .
>
Leave the track and then rejoin in independent mode, then all exercises are
unlocked.
During my testing I've done that multiple times.

samWson:
> aren’t we now divorced from the file system?

@macta I did have my intent wrong with using the filesystem. That helps a lot as that was my number one headache.

> You should be able to iterate over the classes in an exercise sub-pkg/tag and generate the source for each and attach it as a fake “file” (no real file system should be needed).

The idea is to have a string of source code in Tonel format for each class excluding the test class, correct? We were originally doing submitting of the exercises from `ExercismManager>>#submitToExercism:` I had intended to replace the `OSProcess` code with the `ExercismSubmit` command in there. If we continue down that path I can remove the `ExTonelWriter` so we are not writing to the file system anymore.

On generating the source for each exercise 'file': I figure the `ExercismSubmit` command should just get the exercise name as an argument, and from that the command generates the 'files', assembles the HTTP client, and submits the exercise. Let me know if that is what we want, then I can get back to it this weekend.

samWson:
I'm very stuck on this at the moment. I've been trying to get the source code of a exercise in Tonel format before adding it to the submit client as Mime parts (no file system involved, just in the image memory).

My main problem is making a `TonelWriter`. Creating instances seems to be dependent on a source directory. I was hoping in some fashion to create it just using a class or its `RPackageTag` then just get the contents of the writer as a string:

```smalltalk
"Using exercise acronym as an example."
"Trying to get the source code in Tonel format as a string."
packageTag := (RPackageOrganizer default packageNamed: 'Exercism') classTagForClass: Acronym.

"Problem here: a working copy (.mcz file) is expected in the pharo-local\package-cache\"
version := MCVersion package: packageTag. 

"Won't work because of the above issue."
TonelWriter fileOut: version on: (String new writeStream).
```

bencoman:
Having a poke at this myself, I feel your pain.  Tonel doesn't look written to be independent of filesystem.
Not a full answer, but I notice ```TonelWriter-class >> fileOut:on:``` calls...
```
TonelWriter >> writeVersion: aVersion
	self writeSnapshot: aVersion snapshot
```
so better way of getting a snapshot could be...
```
packageTag := (RPackageOrganizer default packageNamed: 'Shout') classTagForClass: SHRange.
snapshot := packageTag snapshot .
```
but then ```writeSnapshot:``` calls...
 ```writePackage:```, ```writeClass:```, ```writeMethodExtensions``` -->```writeExtensionMethods:className:``` 
all of which do... 
```
    self fileUtils 
        writeStreamFor: 
```
which presumes a filesystem and is no use for us to directly get Strings. At which point I see two options:
1. Use a memory file system like...
```TonelWriterTest >> testWriteSnapshot
	| writer mem nl |
	mem := FileSystem memory root.
	writer := TonelWriter on: mem.
	writer writeSnapshot: self mockSnapshot.
```
2. Refactor Tonel methods as our-own-extension** to return a Dictionary of Strings.  The top half of ```writeMethodExtensions``` already does this, so I think this is feasible. As a first pass I'd ignore the ```writePackage:``` and ```writeMethodExtensions```

** Separately we can submit a PR for this to Tonel.

bencoman:
I've been playing around. No time for a PR at 3:30am, but perhaps something like...
```
TonelWriter >> writeClass: aClassDefinition
	[ 
		self fileUtils 
			writeStreamFor: (self fileNameFor: aClassDefinition) 
			in: self packageDir 
			do: [ :aStream | self writeClass: aClassDefinition on: aStream  ] 
	]
	on: TonelShouldIgnore
	do: [ :e | self logCr: 'ignoring: ', aClassDefinition asString ]
```
```
TonelWriter >> writeClass: aClassDefinition on: aStream "extracted from above"
	self writeClassDefinition: aClassDefinition on: aStream.
	self writeClassSideMethodDefinitions: aClassDefinition on: aStream.
	self writeInstanceSideMethodDefinitions: aClassDefinition on: aStream 
```
```
TonelWriter>>mappedSnapshot: aSnapshot
	"extracted from #writeSnapshot:"
	|tonelMap|
	snapshot := aSnapshot.
	tonelMap := Dictionary new.
	(snapshot definitions select: #isClassDefinition)
		do: [ :classdef |  |tonelStream|
			tonelStream := WriteStream on: String new.
	 		self writeClass: classdef on: tonelStream.
	  		tonelMap at: classdef className put: tonelStream contents ].
	^tonelMap
```
```
Playground...
packageTag := (RPackageOrganizer default packageNamed: 'Shout') classTagForClass: SHRange.
snapshot := packageTag snapshot .
tonelMap := TonelWriter new mappedSnapshot: snapshot.
```

samWson:
@bencoman thanks for digging into that. I can understand most of it. It gets a good result. 

We could adopt the methods above as extension methods. We also already have a `ExTonelWriter` so we could override some of the behavior of `TonelWriter` for our use case. I'm in favor of the former since all the tests are still green. If a PR gets accepted its even better for us.

macta:
Nice job moving this forward guys - it shouldn’t have been quite this hard, but the Tonel classes are still quite 1.0 and this usecase probably helps shape them.

Sadly my laptop has died in my travels so I’m even less able to help at the moment - maybe in Sydney I might be able to get it fixed in a few weeks. Anyway, keep pushing.

bencoman:
I'll have a PR for our own extensions shortly.

samWson:
Just a quick update. I have another commit to push using the changes to the TonelWriter shown above, though the TonelWriter changes are not a part of the commit. I'm having trouble pushing things using Git on Windows at the moment. Hopefully I'll get them sorted out by tomorrow.

samWson:
Here is the latest I have on this. Definitely not finished at this stage, but this should be enough for you to see weather I am on the right track or not.

@bencoman this change makes use of the changes to `TonelWriter` you made above, although they are not a part of the commit I made here. One thing I haven't accounted for is iterations of solutions but I think you already posted an example of how to do that on the issue earlier.

There is a random merge commit in there. I find Iceberg far less clear to use compared to Git on the command line, and I ended up making a commit that removed the changes I wanted instead of adding them.

I leave some comments on my thoughts in the code changes for you to see. Still more work to do but hopefully this should be closer.

dev/src/ExercismTools/ExercismManager.class.st
```diff
@@ -72,9 +72,7 @@ ExercismManager >> submitToExercism: packageOrTag [
 				collect: [ :cls | 
 					self pathStringFor: (submissionDirectoryRef / (writer fileNameFor: cls asClassDefinition)) ]).
 				
-	cmd := 'exercism submit ' , filesToSubmit.
-	result := PipeableOSProcess waitForCommand: cmd.
-	result succeeded ifFalse: [ self error: 'Unable to submit solution, CLI reported ', result outputAndError printString ].
+	ExercismSubmit exercise: exerciseName.
```
samWson: I've placed this here as it seems like the easiest way to plug in to what we already have. I don't know if we intend to keep using the `ExercismManager`.

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -1,33 +1,32 @@
...
+		'exercise',
+		'exerciseMap'
+	],
 	#category : #ExercismTools
 }
 
+{ #category : #accessing }
+ExercismSubmit class >> exercise: exerciseId [
+	^ (self new exercise: exerciseId) execute
+]
```
samWson: I've assumed that `exerciseId` is intended to be a string in kebab case e.g. `'hello-world'`. I've never been sure if this is what is expected here.

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -51,3 +50,37 @@ ExercismSubmit class >> solutionDataForExercise: exerciseString [
 		ifNone: [ self error: 'Exercise ''' , exerciseString , ''' not downloaded.'].
 	
 ]
+
+{ #category : #internal }
+ExercismSubmit >> buildRequest [
+	| solutionEntity solutionPart multiPartFormDataEntity solutionId |
+	"Assumption: There is only one entity to make which is from the class with the same name as the exercise."
```
samWson: I've made an assumption as per the comment. Handling a single class with the same name as the exercise might be enough for minimum viable product. I assume we don't need to submit the test class. If needs be I don't think it is too much trouble to handle more than one class either as a part of this change or later on.

macta:
Sorry, I’ll try and have a look tomorrow night. Managed to get my computer fixed in Sydney, so I can better look at code (using an iPhone is too hard to keep the context in my head)

bencoman:
@samWson Sorry for the delay in reviewing this. I've started a new contract that is stretching my capabilities (rerouting a 66kV transmission line). Managed to find time this weekend to have a good crack at it and got it working. I'll pushed the above update.
__________________________________
First I need to correct some of my own previous work...
Running your code I bumped into exercismPackage ensureProperties having
duplicate entries for "a RPackageTag(HellowWorld)" caused by removing the HelloWorld package
then re-running ExercismDownload exercise: 'hello-world'. So rather than being tricky and piggy-backing Pharo infrastructure, I added SolutionData to ExercismManager as a Dictionary.
__________________________________
The CLI submit semantics are...
C:> exercism submit -h
Call the command with the >>>list of files<<< you want to submit.
__________________________________
So even if most exercises so far only need one class, I think in #buildRequest your assumption "There is only one entity to make which is from the class with the same name as the exercise" is too narrow. Our "classes" are more equivalent to "files" and an "exercise" relates more to our "packages". I've reworked it to allow submitting single & multiple classes as well as a whole exercise/package by name.
__________________________________
I found...
ApiPath := '/v1/solutions/latest' was not suitable for SUBMIT since it requires the "/latest" stripped off.
(See the PATCH line under "Now Submitting..." heading far above.)
So I changed that to ApiPath := '/v1/solutions and modified ExercismDownload to match.
IMPORTANT: Evaluate ExercismCommand reset before running #testSubmitHelloWorld*.

dev/src/ExercismTools/ExercismCommand.class.st
```diff
@@ -71,7 +71,7 @@ ExercismCommand class >> configureToken [
 ExercismCommand class >> configureToken: your_CLI_token [
     "Get your_CLI_token at https://exercism.io/my/settings"
     ApiToken := your_CLI_token.
-    ApiPath := '/v1/solutions/latest'.
+    ApiPath := '/v1/solutions'.
```
bencoman: Needed to strip "/latest" to work with SUBMIT.

dev/src/ExercismTools/ExercismDownload.class.st
```diff
@@ -94,16 +95,7 @@ ExercismDownload >> installExercise [
...
-			[	:def | 
-				def load.
-				def isClassDefinition ifTrue: 
-				[	|tag|	
-					"Store solution metadata."
-					tag := exercismPackage classTagNamed:  (exercismPackage toTagName:  def category).
-					exercismPackage ensureProperties at: tag put: solution.
-				].
-			].
+         	parser document do: [:def | def load].
```
bencoman: Solution data now stored in ExercismManager

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -1,53 +1,108 @@
...
 }
 
-{ #category : #accessing }
-ExercismSubmit class >> solutionDataForClass: aClass [ 
-	| exercismPackage classTag |
-	exercismPackage := RPackageOrganizer default 
-		packageNamed: 'Exercism' 
-		ifAbsent: [ self error: 'No exercises downloaded.' ].
-	classTag := exercismPackage classTagForClass: aClass.  
-	^ exercismPackage ensureProperties at: classTag.
```
bencoman: Solution data moved to ExercismManager.

dev/src/ExercismTools/ExercismSubmit.class.st
bencoman: Sounds fine. I've made it a bit more explicit adding asKebabCase is another place.
(p.s. I think the code shifted under this comment, so resolving it.)

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -1,53 +1,108 @@
...
 { #category : #accessing }
-ExercismSubmit class >> solutionDataForExercise: exerciseString [ 
-	| exercismPackage |
-	exercismPackage := RPackageOrganizer default 
-		packageNamed: 'Exercism' 
-		ifAbsent: [ self error: 'No exercises downloaded.' ].
-	exercismPackage ensureProperties 
-		detect: [ :solutionData | ((solutionData at: 'exercise') at: 'id') = exerciseString ] 
-		ifFound: [ :solutionData | ^ solutionData ] 
-		ifNone: [ self error: 'Exercise ''' , exerciseString , ''' not downloaded.'].
```
bencoman: No longer required. Solution data moved to ExercismManager.

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -1,53 +1,108 @@
...
+{ #category : #command }
+ExercismSubmit class >> classes: classes [
+	^ (self new classes: classes) execute
+]
+
+{ #category : #command }
+ExercismSubmit class >> exercise: exerciseId [
+	"Submit whole package-tag specified by exerciseId."
+	"By default, don't submit TestCases with solution (if needed use #classes: direct)"
+	|rootPackage packageTag solutionClasses|
```
bencoman: @samWson I changed you semantics here.  exerciseId to represent a whole package-tag rather than a single class.
Added ```#class:``` for submitting single classes.

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -1,53 +1,108 @@
...
+		solutionEntity := ZnByteArrayEntity bytes: (exerciseMap at: solutionClass name asSymbol).
+		solutionPart := ZnMimePart exercismFieldName: 'files[]' fileName: solutionClass name, '.st' entity: solutionEntity.
+		multiPartFormDataEntity addPart: solutionPart.
+	].
+	
+	solutionId := (ExercismManager solutionData at: packageTag name asKebabCase) at: 'id'.
 	
+	httpclient 
+		url: ApiPath , '/' , solutionId;
+		entity: multiPartFormDataEntity
```
bencoman: Removed  '/latest' from ApiPath to make this work here.

dev/src/ExercismTools/ExercismSubmitTest.class.st
```diff
@@ -10,17 +10,21 @@ ExercismSubmitTest >> setUp [
 ]
 
 { #category : #tests }
-ExercismSubmitTest >> testSolutionDataForClass [
-	|solutionData|
-	ExercismDownload exercise: 'hello-world'.
-	solutionData := ExercismSubmit solutionDataForClass: HelloWorld.
-	self assert: ((solutionData at: 'exercise') at: 'id') equals: 'hello-world'.
```
bencoman: No longer required. Solution data moved to ExercismManager.

dev/src/ExercismTools/String.extension.st
```diff
@@ -13,3 +13,11 @@ String >> asKebabCase [
 						ifTrue: [ kebabStream nextPut: $- ].
 					kebabStream nextPut: char asLowercase ] ]
 ]
+
+{ #category : #'*ExercismTools' }
+String >> kebabAsCamelCase [
+	"Answer a String that converts the CamelCase input to camel-case kebab output 
+	used by exercism"
```
bencoman: Whoops, copy-paste error with the comment, should be...
```suggestion
	"Answer a new String converted from Exercism's kebab-case to CamelCase."
```

bencoman:
On Tue, 23 Oct 2018 at 15:23, Samuel <notifications@github.com> wrote:

> @bencoman <https://github.com/bencoman> thanks for the help. I see in the
> commit message you say that ExercismSubmit is now working, so I take it
> that you have tested this?
>
Yes, I tested in a fresh Pharo 6.1 image.
Note it wasn't a completely fresh download.  A couple of months ago I
downloaded a new Pharo 6.1 image,
updated only Iceberg and archived that as my base for Exercism dev.

I'm having trouble submitting or downloading anything so I'm about to try
> again with a clean image to be sure.
>
You don't mention whether you did ```ExercismCommand reset``` ?
Without that, ```ApiPath``` variable is not updated to the required value,
so both both downloading and submitting will be broken.


> Otherwise I think the only thing left to do is to incorporate the new
> submit code into the Nautilus menu when you right click on an exercise.
>
Once you observe it working,  lets merge this PR.  Then you can add the GUI
part if you like.  I am caught up on other things for a few days.

samWson:
@bencoman thanks. It works now. I'm happy to merge this thing at long last. I'll get started on new issue for adding the new submit code to the GUI part.

macta:
@samWson sorry its been just too difficult to contribute while on the road. I did however try downloading the latest image a few days ago expecting to see it use the work you guys have been forging on with - and it still seemed to be using the OS shell stuff? Possibly user error - but master does have this stuff for both pull and submit right? I did see @bencoman 's improved token prompt but when submitting a solution it seemed to be using os shell. I wanted to confirm its all supposed to be in there, and I'll try again when I get a moment.

bencoman:
I don't think the direct-submit functionality has made it through to the
GUI yet.
Try the (minimal) unit test.

samWson:
@macta we are almost there. Replacing the original code at the GUI is the last bit for this issue. It doesn't look like much work at all, however that's when I ran into my [latest problem with Tonel.](https://github.com/exercism/pharo/issues/96#issuecomment-433596078)

> @samWson sorry its been just too difficult to contribute while on the road.

That's OK. I don't mind real life taking priority over this project any day. Have a good holiday. I'm happy to keep working on the code as I have time available, and I've always got the Pharo Discord available if I can't get help here :)

PR #112, opened 2018-11-07
Issue 96: Submit exercises via GUI

samWson:
Fixes an unknown method on ExTonelWriter, giving it a complete in memory file system for
writing exercise source code.

### DIFF
--- a/dev/src/ExercismTools/ExercismSubmit.class.st
+++ b/dev/src/ExercismTools/ExercismSubmit.class.st
@@ -47,7 +47,7 @@ ExercismSubmit class >> exercise: exerciseId [
 
 { #category : #internal }
 ExercismSubmit >> buildRequest [
-	| solutionClasses multiPartFormDataEntity solutionId exercise |
+	| multiPartFormDataEntity solutionId |
 	multiPartFormDataEntity := ZnMultiPartFormDataEntity new.
 	classes do: [ :solutionClass | |solutionPart solutionEntity|
 		solutionEntity := ZnByteArrayEntity bytes: (exerciseMap at: solutionClass name asSymbol).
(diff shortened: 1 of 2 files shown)


samWson:
I managed to sucessfully submit an exercise from the GUI with the code that uses the http PATCH method from a Windows 10 machine.

It deserves testing from a fresh image to be sure, but I think we can finally call this issue done. I'll leave this PR open for you to review if you have the time, and then I'll merge later.

dev/src/ExercismTools/ExercismSubmit.class.st
```diff
@@ -47,7 +47,7 @@ ExercismSubmit class >> exercise: exerciseId [
 
 { #category : #internal }
 ExercismSubmit >> buildRequest [
-	| solutionClasses multiPartFormDataEntity solutionId exercise |
```
samWson: Just removing unused local variables.

bencoman:
Its cool that you tracked down that problem, but I'm curious why [all this](https://github.com/exercism/pharo/blob/ae8611ae/dev/src/ExercismTools/ExercismManager.class.st#L63-L82) is still needed rather than using ExercismSubmit on its own?  None of that preceding work seems to be used by ExercismSubmit.   
i.e. [similar to this](https://github.com/exercism/pharo/blob/ae8611ae3/dev/src/ExercismTools/ExercismManager.class.st#L38-L49)

Hope you enjoy your break.

samWson:
@bencoman thanks for spotting that. I'll look into it. I'm more than happy to clean up code we don't need anymore.

samWson:
Just pushed a new change to remove the dead code. Thanks for spotting @bencoman.

Issue #83, opened 2018-08-10
I Can Read C++ and Java But I Can’t Read Smalltalk


kytrinyx:
_Imported from https://github.com/exercism/docs/issues/120._
_Originally posted by @bencoman._

#

Should we reference or intermix parts of this document (which I love) ?
[readingSmalltalk.pdf](https://github.com/exercism/docs/files/2276414/readingSmalltalk.pdf)

samWson:
@bencoman Great work. I've been reading the article myself in small bits at time. I think it will be useful.

Bajger:
Article included already in learning resources: https://github.com/exercism/pharo-smalltalk/blob/main/docs/LEARNING.md#starting-out
Issue can be closed.

glennj:
I get nostalgic seeing the name Wilf Lalonde. He taught me Smalltalk at Carleton U almost 30 years ago.

Bajger:
> I get nostalgic seeing the name Wilf Lalonde. He taught me Smalltalk at Carleton U almost 30 years ago.

Lucky you! I remember document from Mr. Lalonde was one of our reference materials, when I studied Smalltalk class back in 2003 😄 
Our study course materials: http://perchta.fit.vutbr.cz:8000/smalltalk
refer to same document: http://www.eli.sdsu.edu/courses/spring02/cs635/readingSmalltalk.pdf

PR #115, opened 2018-11-26
Add exscript from Reading Smalltalk PDF

samWson:
Adding a relevant section from the 'Reading Smalltalk' PDF. I've picked this part as it seems the appropriate place for it.

Let me know if you think this style of using sections of the PDF through out the exercises is the right approach or if we should do something else.

One part of many possible changes for #83.

### DIFF
--- a/exercises/reverse-string/README.md
+++ b/exercises/reverse-string/README.md
@@ -7,8 +7,35 @@ input: "cool"
 output: "looc"
 
 ## Hint
-While there is a #reversed method for Strings, can you figure out how to do this yourself with lower level character iteration or streams? To help you browse code, Pharo has a neat code finder tool. Press Shift-Enter to activate it.NOTE: We have followed the Exercism convention of calling the solution ReverseString, however a more Smalltalk name would be StringReverser.
+While there is a #reversed method for Strings, can you figure out how to do this yourself with lower level character iteration or streams? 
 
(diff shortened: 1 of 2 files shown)


exercises/reverse-string/README.md
```diff
@@ -1,36 +1,63 @@
...
+#thisIsASymbolToo
+```
+
+There are also very few commas evident in Smalltalk programs because they play no syntactic role.
+That’s why array literals, for example, are comma-free; e.g.
+`#(1 2 3 4 5)`
+However, comma is an operator. So you will see it used now and then to concatenate two strings; e.g.
+`'string1', 'string2'`
+Lastly it's worth knowing that strings are collections of characters. This can catch you off guard when
+iterating over strings if you are giving characters to a method that expects strings.
```
samWson: I've added my own comment here about strings and characters as I know from my own experience I have been confused when I handed a method a character thinking it was a string.

samWson:
I've added a link to the entire PDF on this repositories README. I figure it should be easily accessed in full at least somewhere and the contributing section is the best part.

I'll stick in more sections in exercise READMEs as appropriate.

bencoman:
> Let me know if you think this style of using sections of the PDF through out the exercises is the right approach

That was the idea.  I was hoping to retain the sequence of the PDF and matches exercises into it, but I hadn't looked into that yet to know if that even makes sense.

PR #116, opened 2018-11-29
Add section on collections from Reading Smalltalk PDF

samWson:
I figure since our solution for sum-of-multiples uses collections and iteration, this is a good place for this guide on collections. Taken again from the reading smalltalk PDF.

### DIFF
--- a/exercises/reverse-string/README.md
+++ b/exercises/reverse-string/README.md
@@ -13,30 +13,6 @@ To help you browse code, Pharo has a neat code finder tool. Press Shift-Enter to
 
 NOTE: We have followed the Exercism convention of calling the solution ReverseString, however a more Smalltalk name would be StringReverser.
 
-## Smalltalk and Strings
-
-There are some obvious little details like double quotes for comments, single quotes for strings, and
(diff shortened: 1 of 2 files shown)


samWson:
Hold on merging this for now. I think I've gone about this the wrong way. I've been adding to the exercise READMEs directly on Github, but now I see they are supposed to be generated from templates.

I'll look into generating these properly and undoing my manual edits to the READMEs.

samWson:
@macta yes I'm going through that now. Looks like the right way for me to do this would be to add the text I want to the test class comment of an exercise then generate the README from there.

samWson:
Just added a couple of commits to undo the manual edits I made to the READMEs. Instead I will use the removed text with the configlet tool to generate the READMEs properly.

macta:
@samWson As I was updating the comments of the existing tests to match what the exercise generator does - I've applied your fixes in there. So I think this PR can be deleted when we approve that one.

macta:
This is now obsolete as those comments were merged into the revlevent test comments. The exercise creation process is much easier now too.

PR #118, opened 2018-12-09
pharo7-clean-load

bencoman:
Move menu pragmas off AbstractNautilusUI so packages can load cleanly in Pharo 7.


bencoman:
I was having some issues with Iceberg in Pharo 6.1 and popped in to load Exercism in Pharo 7.
Pragma's can be put on any class.  Their purpose is to reduce class dependencies.

Issue #124, opened 2019-02-15
The Welcome example is smacked by the code critic for not being in a Test package


macta:
When you load up exercism - the Welcome test isn't like by code critic and shows 2 warning signs which would put off a new user.

Either need to:
1. disable the critic for those (I think there is some meta info you save), 
2. or we turn off critic  in the load setup, 
3. or we fix it like critic wants.

Possibly the 1st is best, as this is just a little warmup thing for users and we're trying to sidestep too much info upfront.

macta:
This just needs retesting in a fresh install and then it can be closed

macta:
Its fixed

PR #126, opened 2019-02-15
124 (cherry pick) The welcome example is smacked by the code critic

macta:
@bencoman your git fu is stronger than mine - I can't explain why this PR seems to drag in the 3 other commits that I did last night? Which I thought were already on master?

I tried going back to master (which has those 3 commits), creating a new branch (124-Cherry-pick) and then I did a cherry pick to pull out the classes I changed.

I expected then just see a branch with those changes, but it still keeps showing me the original 3 commits - and github warns it can't automatically merge? Do I need to delete those other branches first?

@samWson - if I can do this cleanly, it will address #124 quite cleanly.


macta:
@bencoman actually maybe I was reading this wrong (6 months on the road has got me out of practice) - the actual files changed is only 4 (what I intended). I'm still not clear why the other commits are dragged in - particuarly if I cherry picked (I thought that was the intent of cherry picking - but maybe I've misunderstood that, and it drags in the other commits to let you know where you picked from?)

macta:
@bencoman I think I've seen what I've done - I've not resynced my fork with exercism/pharo and so I'm branching off an earlier point (Doh). Need to refresh my memory and how to do that again.

bencoman:
@macta Not sure if this directly addresses your issue, but btw, once you have cloned from macta/pharo-exercism to can add exercism/pharo as a remote and fetch/push directly to there rather than via macta/pharo-exercism (and maybe you are already doing that)

bencoman:
> Do I need to delete those other branches first?

I believe "other branches" should never get in the way.  In some respects "branches" are really unimportant. Branches are just a movable "pointer" to a particular commit.  

But I have experienced similar before with old pharo-project/pharo repos.  Seems something to do with finding the last common commit or which direction the merge is being done, but I haven't been able to put my finger on it.

macta:
@bencoman as you helpfully commented in the Pharo newsgroup - all this mess was down to me not merging my fork downstream. I'm slowly figuring this out, but if you get off piste, its tricky coercing everything back in line.... I'm slowly doing that, and will redo this pr given my new knowlege

PR #133, opened 2019-02-18
Update instructions for new repo-name that doesn't conflict with base…

macta:
… pharo development

Slightly trim the example code snipped so it fits in the newer panel design


macta:
Applying so we are good for more testing

PR #135, opened 2019-02-19
Improve the menu's for exercism by cleaning up to use instance side methods

macta:
… improve the ordering and provide a fetch menu at both the exercise and project level (so its less annoying to click on the right place.  *we may change fetch on the exercise level to fetch next as we could have the meta info in the image).

(cherry picked from commit 5a0b684)

This PR is a second attempt to do this cleanly due to forked repo shenanigans


macta:
@samWson I would apply this PR before testing as it makes it a bit more usable - but I wanted to get your view on it first.

macta:
As this is not very controversial and tiny, I'm going to merge and you guys can feedback later. I'd like to clean up some of my branching mess caused by my Fork having some commits on master (I think)

PR #136, opened 2019-02-19
Adds generator menu commands and updates it to work with the latest Tonel

macta:
(cherry picked from commit ba7242d)

This PR fixes some fork shenanigans and gives a clean diff of the code required.

### DIFF
--- a/dev/src/ExercismDev/ClyExercismGenerateCommand.class.st
+++ b/dev/src/ExercismDev/ClyExercismGenerateCommand.class.st
@@ -1,0 +1,35 @@
+Class {
+	#name : #ClyExercismGenerateCommand,
+	#superclass : #ClyExercismCommand,
+	#category : #'ExercismDev-Menus'
+}
+
+{ #category : #testing }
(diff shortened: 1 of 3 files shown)


macta:
@samWson This PR fixes the generation side of development - and answers your older question about what was the sourceDir: error you were seeing. The TonelWriter isn't very generic - but it seems that it got updated a bit so there is a new hook to specify where code gets written - we can override that objectXXXX method and use it to control where code gets written.  I still think TonelWriter could be a bit more configurable - but our subclass is a bit less nasty now (and so shouldn't break as we don't need to subvert an internal instance variable)

macta:
@samWson Its this generator you have to use to run the exercism configlet to regenerate the templates for the exercise documentation (what you did wrong when trying to update the ReverseString docs).

We might consider at some point generating ourselves - but not sure if its worth the effort yet - we need more exercises. Also as configlet runs in their build process - we need to make sure it works cleanly.

macta:
@samWson, as I'd like to pair with a mentee on improving code and would like to use Exercism as a handy example, I'm going to merge this to reduce any code conflict issues.

PR #139, opened 2019-02-20
This cleanup better factors the menu commands (can use a second metho…

macta:
…d with annotation for fetching vs. having a second class).

Also moves the debug meta option to the dev package (so not in user mode)


macta:
This one has been sitting for review, and I think you guys are all busy, so I'm going to push it.

PR #143, opened 2019-02-22
Implement `space-age`.

jaccarmac:


jaccarmac:
I've finished the existing exercises and would like to start contributing (exercises and scaffolding as I get more familiar with it). Ported the `space-age` exercise over into a new dev image. As far as next steps to get this PR out of draft status, I figure I need to.

- [] Write documentation for most/all of my classes/methods.
- [] Generate the appropriate files for Exercism. (I gather this is by using the built-in example method but am not quite sure.)
- [] Make sure my class-method based solution isn't horribly unidiomatic.

The first two I can fiddle with on my own, but I am especially open to pointers about the second. Right now the repository is just checked out in Iceberg, pointing to my fork as well. To generate files I guess I need to have `configlet` on my path and a version of the specs sitting next to `pharo-smalltalk`. Is this usually done in the Iceberg directory or on a pristine copy of the repo somewhere else on disk?

Any seasoned Smalltalkers want to point out where I erred on the last point?

jaccarmac:
Seems like adding a `.gitattributes` might help solve some line ending woes.

macta:
Thanks for getting involved - I'll have a look on the weekend.
I need some help with this as there are a lot of exercises. In fact I've just used the https://github.com/exercism/problem-specifications to generate stubs for all the remaining ones.
(SpaceAge was in that list).

There are a bunch of config things you have to do to get them into Exercism and its not so obvious - but by generating I'm getting close to automating that bit too - and then we just need to actually solve the problems (and rationalised what I generated from the schema). 

Anyway - if you are interested and want to help - thats great. I'll see what you've done so far.

dev/src/Exercism/Planet.class.st
```diff
@@ -0,0 +1,20 @@
...
+I represent facts about planets
+"
+Class {
+	#name : #Planet,
+	#superclass : #Object,
+	#category : #'Exercism-SpaceAge'
+}
+
+{ #category : #testing }
+Planet class >> isAbstract [
```
macta: Nice touch on implementing isAbstract properly (this has caught me out numerous times)

macta: While I'm a big fan of objects for everything - I don't think this hierarchy is necesssary for what is effectively a list of constants. All these classes implement a constant #orbitalPeriod and don't have any behaviour.

I would just have dictionary of constants - and retrieve them until there was something more significant to reify them (and this is slightly driven by the nature of the Exercism exercises that are normally aimed at just having 1 or 2 classes to make sharing solutions online a bit easier)

dev/src/Exercism/SpaceAgeTest.class.st
```diff
@@ -0,0 +1,61 @@
...
+
+{ #category : #asserting }
+SpaceAgeTest >> assert: actualNumber closeEnoughTo: expectedNumber [
+	"The canonical data is precise to two decimal points. Number >> closeTo:precision: exists but didn't give quite the correct behavior."
+
+	self
+		assert: (actualNumber roundTo: 0.01)
+		equals: (expectedNumber roundTo: 0.01)
+]
+
```
macta: Nice spot on this one - I'd like to double check on the mailing list as we should be able to use something out of the box? But good refactoring usage

exercises/acronym/README.md
```diff
@@ -8,7 +8,7 @@ Help generate some jargon by writing a program that converts a long name
 like Portable Network Graphics to its acronym (PNG).
 
 ## Hint
-You need to find a way to split up a String and then iterate cleanly over the resulting sections to output a result. 
+You need to find a way to split up a String and then iterate cleanly over the resulting sections to output a result.
```
macta: When submitting - check your results, as you didn't knowingly change this so it shouldn't be part of your commit?

exercises/reverse-string/README.md
```diff
@@ -1,63 +1,36 @@
...
+For example:
+input: "cool"
+output: "looc"
+
+## Hint
+While there is a #reversed method for Strings, can you figure out how to do this yourself with lower level character iteration or streams? 

To help you browse code, Pharo has a neat code finder tool. Press Shift-Enter to activate it.

NOTE: We have followed the Exercism convention of calling the solution ReverseString, however a more Smalltalk name would be StringReverser.
```
macta: Again - you shouldn't be sumbitting this - however it does show your comment about the .gitattributes suggestion, which we need to look into (I presume this comes from windows - and using the configlet generatation tool on that platform?)

exercises/reverse-string/README.md
```diff
@@ -1,63 +1,36 @@
...
+If you are still stuck, the track documentation has more detailed help on [running tests](https://exercism.io/tracks/pharo/tests).
+
+## Language and Environment Help
+
+For Pharo installation and learning resources, refer to the [track help page](https://exercism.io/tracks/pharo/learning).
+
+
+## Source
+
+Introductory challenge to reverse an input string [https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb](https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb)
```
macta: This huge change is mistake that @samWson made in not understanding the template generation through configlet.

I'm actually thinking (now we can generate the tests from the specifications) that I might include the full comment in the test comment. More to come on this if you are interested (I'll include you in my PR on monday)

dev/src/Exercism/Planet.class.st
jaccarmac: Doubted this approach the more I thought about it, just felt right at first. I think it could be an interesting way to approach the problem if the classes already existed in the image, and an opportunity to teach about packaging/extension methods/protocols in Pharo. But, indeed overkill for the exercise. I'll change it once the stub patch drops.

dev/src/Exercism/SpaceAgeTest.class.st
jaccarmac: `closeTo:` exists but didn't quite do what I wanted. Out of the box would be ideal, let me know what you find.

exercises/acronym/README.md
jaccarmac: This and some other stuff was part of stuff configlet changed and was in a separate commit. I can rip it out.

exercises/reverse-string/README.md
jaccarmac: I'm on Linux but same difference really. And the issue seems to be caused by Pharo, not the OS. Git deals with line endings differently based on your global config and Windows is particularly pathological. It's likely that different committers have different combinations of line ending conversion settings in Git. My opinion is rather against doing conversion, but happily we can specify rules for specific files in `.gitattributes`. In the `README.md` case I would imagine `^M` should turn into a Unix line ending?

exercises/reverse-string/README.md
jaccarmac: :+1: I'll look for that PR.

dev/src/Exercism/Planet.class.st
bencoman: Classes are cheap in Pharo.  I think it would be good to push such a class-based solution at some point, since its a point of distinction in approach to object-oriented. But in doing so we need to avoid Pharo appearing overly complicated. So maybe later and maybe need to find the right exercise.  

Actually it would be good to be able to distinguish between default simplest solution and class-based solution as an extension.   Perhaps the latter could be an extension exercise named `space-age-OO` - but not right now.

macta: As @bencoman mentions - yes we shouldn't be scared of classes (and Smalltalk favours using classes even if they only have a few methods in them), but I haven't studied some of the later exercises enough to see if they develop the idea that classes are ok (as this is often the flaw in training courses like exercism) - however the Planets don't feel meaty enough in this example, and the problem doesn't lead you to "double dispatch" anything further on to them. So I think they really are just constants. I'd be interested in seeing what a constant's solution would look like (which is probably a static dictionary I'm guessing).

macta: @jaccarmac I hope the feedback didn't come accross too harsh - we are feeling along with you on this too.

exercises/reverse-string/README.md
macta: I'm not so familiar with .gitattributes (just had a look) - is it simply a case of putting:
```
*.md text
```
in a file .gitAttributes (I think its the only thing that seemed to get upset as Pharo was handling the rest ok).?

dev/src/Exercism/Planet.class.st
jaccarmac: Not at all. I'm treating GitHub as a poor man's mentored track :)

exercises/reverse-string/README.md
jaccarmac: That or `text eol=crlf` would work I believe.

exercises/reverse-string/README.md
macta: @jaccarmac - I've just pushed PR #144 (its a bit big) but If you read the pr text, and look at the examples - I'd be interested in your take. And possibly we can use that generated example as a base for your SpaceAge?

macta: @jaccarmac - I've now pushed all the generated solutions (SpaceAge is one of them) - lets see how well the readme instructions/flow work out (and you can hopefully help me tune it)

macta: @jaccarmac we've slightly adjusted the format to use numbered tests as this is what the exercism spec calls for (at least ordered tests) - you should be able to pull space age out from this.

jaccarmac: Yeah I managed to turn my image into a bit of a mess and haven't gotten around to recreating it yet. Will with the new code for sure.

PR #144, opened 2019-02-25
Generate exercises from spec

macta:
This is a big commit (sorry, I should have chopped it up better but got excited)
This is a bit of a spike, but in the interest of forging on - I am proposing to push it out and then write some tests around it (to be honest, its pretty boiler plate, non fancy code)

The idea is that if you checkout the exercism problem-specifications.git in a terminal (to a known location), then in Pharo you click on the  #generate method of ExercismExerciseGenerator, which will prompt you for the exercises directory of the git checkout and then will iterate through all of the examples, parsing the specification and generating a test class and test methods into an ExercismWIP package. (you only need to do this once, and then afterwards when upstream exercises are added).

I've included 2 examples (Hamming and SpaceAge) so you can see what they look like. If we like it, I propose to generate the remaining 103 examples.

The exercise tests, also have some class methods (which I've also pushed up into an abstract ExercismTest class) that give us some meta information like, version, uuid and the next test - which we can use to generate the config.json file automatically (future story). 

I think all of this then makes it easier for people to contribute as we then simply ask them to:

- check out the development branch
- find an exercise they like in the WIP package
- TDD a solution and adjust the tests accordingly (the generator is pretty basic - this said, corrections might be appropriate as PR's back to the upstream problem-specification text)
- update the meta data (difficulty, unlockedBy, followedBy, Hint text etc).
- get it reviewed with us
- update the package of the example to Exercism (i.e. move it out of WIP)
- run the generator to create the Exercism readme (via configlet)

I think this will put us much further ahead - as we already have some guys itching to do more exercises and they can use this infrastructure/process to help. I'd also like to try and pitch it at PharoDays to get 50+ exercises completed by the community.

### DIFF
--- a/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
+++ b/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
@@ -12,12 +12,14 @@ BaselineOfExercism >> baseline: spec [
 		for: #common
 		do: [ self setUpDependencies: spec.
 			spec
-				package: 'Exercism'
-					with: [ ];
-				package: 'ExercismDev'
-					with: [ spec requires: #('Exercism' 'ExercismTools')];
 				package: 'ExercismTools'
 					with: [ spec requires: #('OSProcess' 'ExternalWebBrowser') ];
+				package: 'Exercism'
+					with: [ "Standalone, as this has all the solutons as well as tests"];
+				package: 'ExercismWIP'
+					with: [ "Holding package for WIP tests"];
+				package: 'ExercismDev'
+					with: [ spec requires: #('ExercismTools' 'Exercism' 'ExercismWIP')];
 				group: 'default'
 					with:
 					#('ExercismTools');
--- a/dev/src/Exercism/Acronym.class.st
+++ b/dev/src/Exercism/Acronym.class.st
@@ -7,11 +7,14 @@ Class {
 	#category : #'Exercism-Acronym'
 }
 
-{ #category : #example }
-Acronym >> abbreviate: aString [
+{ #category : #exercism }
+Acronym >> abbreviatePhrase: aString [
 	| words |
-	words := aString splitOn: [ :char | char isAlphaNumeric not ].
+	
+	words := aString splitOn: [ :char | char isAlphaNumeric not and: [ char ~= $' ]].
 	^ String
-		streamContents:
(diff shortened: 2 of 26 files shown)


jaccarmac:
Like the look of the diff, will have to pull and give the classes a play in an image later today. :+1:

bencoman:
go! go!

jaccarmac:
Like it a lot. The test methods are easy enough to rename and it seems like a pretty foolproof way to scaffold out the tests and metadata (which might otherwise be missed). I'll rebase against it if it gets merged.

macta:
fixed by #149 (the merge suggested by github)

bencoman:
@macta Just had a chance to review this a bit more.  I'm not clear, but this seems to just generate the exercises and metadata, not the actual tests.  Could you clarify your plan for that?
For example I was looking at the TwoFer test invoking method `who:` 

```
TwoFerTest >> testANameGiven [
	self assert: (twoFer who: 'Alice') equals: 'One for Alice, one for me.'
]
```
https://github.com/exercism/pharo-smalltalk/blob/master/dev/src/Exercism/TwoFerTest.class.st#L97

and wondered how that was derived from `Name` parameter of the canonical data...

```
    {
      "description": "a name given",
      "property": "twoFer",
      "input": {
        "name": "Alice"
      },
      "expected": "One for Alice, one for me."
    }
```
https://github.com/exercism/problem-specifications/blob/master/exercises/two-fer/canonical-data.json#L17

macta:
@bencoman actually TwoFer is a bad example as it predates all of this work... look at SpaceAge or Hamming which were generated that way.  

If you recall (but perhaps not obviously) that we discussed leaving Hello and TwoFer as more idiomatic Smalltalk (the former and first exercise avoiding the difference between class and instance - just like main() in java/c) and the latter not having the numbered test methods "test01_" etc. Following these 2 tests (which are very trivial) its all generated via ExercismExerciseGenerator>generate

In retrospect, we should note that in those tests - as its useful info about the journey we wanted to take students on.

Your comment also highlights where I've been naughty - generation was a spike (albeit a successful one) and so guiltily doesn't have the level of testing it should have.  As I was keen to just get tests generated to stay slightly ahead of @jaccarmac (and others) who wanted to contribute, I pushed go. But I think now we understand that re-generation (#170) is actually a thing as the tests will keep getting refined, and so we need proper testing (#174)

bencoman:
Don't mistake my curiousity for implied guilt :).
its fine to move fast and clean up soon after.
Thanks for the time-sequence info.

PR #145, opened 2019-02-25
Fix the abstract menus to use the proper isAbstract convention

macta:
This is a simple tidy up, as I had misunderstood how #isAbstract worked on the class side


macta:
closing this as its a trivial cleanup

PR #146, opened 2019-02-25
Fix readme for pharo7

macta:
Some small housekeeping changes based on feedback from #143


macta:
Now pr #144 is in master, the readme is correct

Issue #148, opened 2019-02-26
Sort out migrating Exercism pre-req projects to Tonel and Runtime only


macta:
We depend on PharoWebBrowser (to cross platform launch a browser to view exercism progress). Its a nice solution however its still in filetree format and causes error on windows ( #140 ).

We need to migrate this library and pre-reqs to Tonel (in part this can be slightly short circuited by specifying a better baseline which has a group directive to let you load only the runtime requirements and not the supporting tests which tend to have the long filenames).

Issue #140, opened 2019-02-21
64 bit windows installation is flakey - gives Git Errors 


macta:
Started testing again from scratch and got an error on 64bit windows:
```
LGit_GIT_EEXISTS: Failed to stat file 'C:/Users/sam/Documents/Pharo/images/exercism-test-run-7.0-32...stance/testRaiseUnexpectedMessageWhenTooMuchMessages.st': The filename or extension is too long.
```

I recall we have had this problem before: filenames too long when running on Windows? I tried to look for another issue where we had this problem for a fix but I haven't been able to find one.

_Originally posted by @samWson in https://github.com/exercism/pharo-smalltalk/issues/120#issuecomment-465902340_

On Discord/exercism-dev @roelof reports a similar kind of problem on Wndows 10:

He Was able to install Exercism via the playground the first time on 64bit windows - but then wasn't able to submit his hello-world exercise (he got a 404 error) - see separate issue.

He then tried on 32bit windows and it was fine.

When he tried installing 64bit a second time he got the long filename error with filename:
```ruleRBUtlitiesMethodsRuleV1FalsePositve.st```

**Comment by RoelofWobben** [comment]
here a screenshot of the problem 

![afbeelding](https://user-images.githubusercontent.com/47429430/53168084-2025c680-35da-11e9-8f79-19a1fe5e54a2.png)

**Comment by macta** [comment]
We've now identified this is also an issue with 32bit as well (which makes more sense).

The workaround is to use the shared repository setting in Iceberg, and set it to a short path like c:\repo

This still raises the question why it fails on files that aren't even in our project.

**Comment by macta** [comment]
@samWson It would be handy to get the full path of the error file you get above (e.g. testRaiseUnexpectedMessageWhenTooMuchMessages.st) - as I think there might be a deeper iceberg bug going on.

Can you do it and get a longer path to see if it is in one of our pre-reqs?

**Comment by samWson** [comment]
@macta Here is the error message:
```
'C:/Users/sam/Documents/Pharo/images/exercism-test-run-7.0-32bit/pharo-local/iceberg/dionisiydk/StateSpecs/StateSpecs-DSL-PragmaBasedShouldExpressions-Tests.package/SpecOfShouldExpressionDriverTests.class/instance/testRaiseUnexpectedMessageWhenTooMuchMessages.st': The filename or extension is too long.
```

and here is the file path it refers to:
```
[...]

PR #150, opened 2019-02-26
Removes unneeded external dependency as there is a WebBrowser access …

macta:
…in the core image

Move OSProcess as a dev only dependency as its only now needed for generating meta data via configlet


macta:
@samWson @bencoman Hurrah - this should be solution to #140 as it was pointed out that there is a WebBrowser class in core pharo. I've tested it in a clean image, so am going to push it so some of our discord users can give it a spin. But Sam - I'm keen for you to then test #140 .

macta:
fixes #148 (by removing the pre-req)

samWson:
@macta I've tested on Windows 10 from scratch again and it seems to be working fine.

macta:
Hurrah!!!!!!!

PR #151, opened 2019-02-26
Generate the remaining exercises so that the community can begin to h…

macta:
…elp writing solutions

This is a big one! However this is all generated so we have a starting point to write solutions.

Contributors can now follow the steps outlined in the readme.md for contribution

### DIFF
--- a/dev/src/Exercism/Acronym.class.st
+++ b/dev/src/Exercism/Acronym.class.st
@@ -11,7 +11,7 @@ Class {
 Acronym >> abbreviatePhrase: aString [
 	| words |
 	
-	words := aString splitOn: [ :char | char isAlphaNumeric not and: [ char ~= $' ]].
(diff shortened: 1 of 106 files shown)


macta:
I'm commiting this one to unblock @jaccarmac

jaccarmac:
Hey re: this. Is there an easy way to change tags for a bunch of classes? I see a context menu item to change packages but then I have to edit tags manually which is annoying because it changes the tag currently being viewed.

macta:
> Hey re: this. Is there an easy way to change tags for a bunch of classes? I see a context menu item to change packages but then I have to edit tags manually which is annoying because it changes the tag currently being viewed.

You can multiple select classes in a package and then use the Move command in the context menu (meta-MC). Although it seems like you can't sub-tag them (thats a bug/missing functionality I think)

PR #152, opened 2019-02-26
Scan solutions for known selectors that the student was asked to avoid

### LINKED ISSUE #142: Add a reverse test to stop users simply using #reverse
This is actually a generic issue, but a good concrete example is to detect in the Reverse exercise if a user has used the selector #reverse and so hasn't learned as much. We could apply this to other exercises as well (maybe generically provide a list of selectors that shouldn't be used)

macta:
This fixes #142 and provides a generic way for tests to scan for misused selectors.

### DIFF
--- a/dev/src/Exercism/LeapTest.class.st
+++ b/dev/src/Exercism/LeapTest.class.st
@@ -89,6 +89,11 @@ LeapTest >> setUp [
 ]
 
 { #category : #test }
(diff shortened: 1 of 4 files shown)


macta:
@bencoman @samWson what do you guys think? This came about when a student on discord was asking about cheating and it occured to me that Smalltalk is very good at this.

PR #153, opened 2019-02-26
Regenerate exercise documentation for new style exercises

macta:


macta:
This was incorrectly generated. ignore

PR #154, opened 2019-02-26
Regenerate the exercise instructions based on the new exercise format

macta:
Needed to run the generation to match the editted exercises


macta:
Not sure if this will get built yet on our track as they are still investigating the changes needed to rename our repo.

PR #159, opened 2019-02-27
Make the config meta info consistent

macta:


macta:
Make the test meta info consistent (not used yet)

PR #161, opened 2019-02-27
Raindrops exercise

### LINKED ISSUE #160: Ordering tests
Reading this...
> However, in order to simulate TDD we want tests to run _in the order that they are defined_, and we want them to _fail fast_, that is to say, as soon as the test suite encounters a failure, we want the execution to stop.

@bencoman thought: One approach would be a downloaded exercise having test methods named:
```
ExceriseTest>>01testSomething
ExerciseTest>>02testSomethingElse
ExerciseTest>>03testAnotherThing
```

We currently order the tests by the name as in the problem-specification (however they are listed in a specific order in that definition, which hadn't occurred to me).

How important do we deem this requirement?

You can of course click on the tests one by one - but the suggestion of prefixing a number is a good one (albeit a bit ugly - but the best we have)

macta:
New exercise - also follows new test naming convention that puts the tests in increasing order of difficulty

fixes #160


bencoman:
The thing with naming like "test18_TheSoundFor3125IsPlangAsItHasAFactor5" is that the test is immediately active.  My idea (not sure if it was the best way) was to have the later tests initially disabled by naming them like "18test_TheSoundFor3125IsPlangAsItHasAFactor5" with the student renaming the test (via a helper  context menu item) to enable it.  I interested in your thoughts comparing these two ways of managing the tests.  I guess with the later its less work for the user to just click the run-test-icon one at a time in order.  Does TestRunner display failed tests in order?

macta:
Hi Ben - in practice, you get a grey dot next to each one, and you can work your way down clicking on a dot and running it one by one manually (without fussing about renaming things). The runner will then run them all (although not sure if the order is guaranteed - but we probably could as they are now alphabetical).

However in reality I think as they are so easy to run this way - it works out fine and I doubt many will run all until they've clicked a few dots and got something running.

macta:
As they are displayed in alpha order you will see red and green dots in that ascending order (so it you run them all you would get some initial green dots up to where you've solved the problem - whether they actually ran in that order I'm not sure, but I don't think you would notice)

macta:
Happy to push this one and try it out?

macta:
I agree on the note to show it’s not normal - I thought that too... but I have seen a test candidate struggle and he started off with a non simple test.

I also thought that maybe the first couple wouldn’t have that convention - and then we can explain it in the note on the first one that does (eg hello world and two-fee don’t need it)

PR #164, opened 2019-02-27
Update existing tests to new number format

macta:
Fix generator to handle numbering format for nested cases


macta:
We are now getting close to having something that is uniform

PR #166, opened 2019-02-28
Implement SpaceAge from the new package.

jaccarmac:
Continuation of #143, which was accidentally deleted during branch rename fun times.

### DIFF
--- a/dev/src/Exercism/SpaceAge.class.st
+++ b/dev/src/Exercism/SpaceAge.class.st
@@ -1,0 +1,45 @@
+"
+I return ages in planet-years for planets in the solar system.
+"
+Class {
+	#name : #SpaceAge,
+	#superclass : #Object,
+	#classVars : [
+		'OrbitalPeriods'
+	],
+	#category : #'Exercism-SpaceAge'
+}
+
+{ #category : #'class initialization' }
+SpaceAge class >> initialize [
+	self initializeOrbitalPeriods
+]
+
+{ #category : #initialization }
+SpaceAge class >> initializeOrbitalPeriods [
+	"Initialize orbital periods in seconds for each solar system planet"
+
+	| earthOrbitalPeriod |
+	earthOrbitalPeriod := 31557600.
+	OrbitalPeriods := Dictionary new
+		at: 'Mercury' put: earthOrbitalPeriod * 0.2408467;
+		at: 'Venus' put: earthOrbitalPeriod * 0.61519726;
+		at: 'Earth' put: earthOrbitalPeriod;
+		at: 'Mars' put: earthOrbitalPeriod * 1.8808158;
+		at: 'Jupiter' put: earthOrbitalPeriod * 11.862615;
+		at: 'Saturn' put: earthOrbitalPeriod * 29.447498;
+		at: 'Uranus' put: earthOrbitalPeriod * 84.016846;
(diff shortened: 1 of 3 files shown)


jaccarmac:
@macta if you would be so kind as to take another look :)

dev/src/Exercism/SpaceAge.class.st
```diff
@@ -0,0 +1,45 @@
...
+	self initializeOrbitalPeriods
+]
+
+{ #category : #initialization }
+SpaceAge class >> initializeOrbitalPeriods [
+	"Initialize orbital periods in seconds for each solar system planet"
+
+	| earthOrbitalPeriod |
+	earthOrbitalPeriod := 31557600.
+	OrbitalPeriods := Dictionary new
```
macta: This approach is much more compact.

Trying this out for an idea - to show that Pharo is not hugely different from langagues like JavaScript - what if we used:
```
:= { 'Mercury' -> earthOrbitalPeriod * 0.2408467.
'Venus' -> earthOrbitalPeriod * 0.61519726.
...
} asDictionary.
```
Not totally sure myself yet - but thought I would ask.

dev/src/Exercism/SpaceAgeTest.class.st
```diff
@@ -52,63 +67,55 @@ SpaceAgeTest >> setUp [
 { #category : #tests }
 SpaceAgeTest >> test01_AgeOnEarth [
 	| result |
-
-	result := spaceAgeCalculator agePlanet: 'Earth' seconds: 1000000000 .
-	self assert: result equals: 31.69
+	result := spaceAgeCalculator ageOn: 'Earth' at: 1000000000.
```
macta: could you refactor rename - ageOn: to ageOnPlanet:

I ask this because I'd like to keep in line with the problem specifications that we generate from - they specify like this:
```
 {
      "description": "age on Earth",
      "property": "age",
      "input": {
        "planet": "Earth",
        "seconds": 1000000000
      },
      "expected": 31.69
    },
```

So I concatenated `age` and `planet` - however I like your idiomatic `on` - making me think I should store some meta information on tests we convert saying something like:
```
inputPrefixes
  ^{ 'planet' -> 'On' } asDictionary

So that we could easily regnerate again if the spec changes.

dev/src/Exercism/SpaceAge.class.st
jaccarmac: I didn't know about `Association` myself, thanks! Updated.

dev/src/Exercism/SpaceAgeTest.class.st
jaccarmac: Done. Not sure how much this could be automated in the future, but would be cool if it could. Changing the message names toe be more natural seems natural for implementing problems to me, I guess.

macta:
I am pulling this into a branch to see it... not sure how to show that here?

macta:
Not sure if you can see this pr in a branch #172 - but it looks fab (love the hint comment)

jaccarmac:
Cool, thanks. I managed to load your PR last week by adding your remote in Iceberg, but that's too annoying for regular usage. Already have some other annoyances with Iceberg re: remotes so might take a look at adding the ability to grab pull requests.

macta:
Closing this, as it was left over from when we had to move this to a new branch.

Issue #42, opened 2018-08-04
Exercise generator should also write out to config.json


macta:
When generating the exercises, the generator could/should also generate the config.json so that exercises are properly linked to exercism and it doesn't require an extra manual step to remember.

It actually would be useful if the exercise package was annotated (possibly via the testcase) with the uuid of the exercise and other meta information needed to create a config.json entry.

The uuid, could also be handy to allow pharo to open the correct web page for an exercise (its based on uuid)

macta:
PR #144 introduces some basic meta info that can be used for this (e.g. uuid, difficulty and followedBy)

PR #167, opened 2019-02-28
Auto generate config files

macta:
An sort of spike to generate the required json config file for the exercises.


macta:
Fixes #42 and should bring us to an easy place to integrate changes.

PR #172, opened 2019-02-28
Jaccarmac space age master

macta:
Moving @jaccarmac's pr to a branch to stage it up


dev/src/Exercism/SpaceAge.class.st
```diff
@@ -0,0 +1,32 @@
...
+	earthOrbitalPeriod := 31557600.
+	^ {('Earth' -> earthOrbitalPeriod).
+	('Mercury' -> (earthOrbitalPeriod * 0.2408467)).
+	('Venus' -> (earthOrbitalPeriod * 0.61519726)).
+	('Mars' -> (earthOrbitalPeriod * 1.8808158)).
+	('Jupiter' -> (earthOrbitalPeriod * 11.862615)).
+	('Saturn' -> (earthOrbitalPeriod * 29.447498)).
+	('Uranus' -> (earthOrbitalPeriod * 84.016846)).
+	('Neptune' -> (earthOrbitalPeriod * 164.79132))} asDictionary
+		at: aPlanet
```
bencoman: Two things occur for me here:
1. Don't repeat yourself
2. The dictionary is created again each time the method is called.  I see an earlier commit created the dictionary in the class-initialisation.  I curious why the change away from that?
I'd lean towards something like...
```
SpaceAge class >> initializeOrbitalPeriods
   RelativeOrbitalPeriod := Dictionary newFromParis:  #(
        'Earth'       1.0 
	'Mercury'     0.2408467
	'Venus'       0.61519726
	'Mars'        1.8808158
	'Jupiter'    11.862615
	'Saturn'     29.447498
	'Uranus'     84.016846
	'Neptune'   164.79132
    ).

SpaceAge >> orbitalPeriodOn: aPlanet
     "Earth orbital period is 31557600 seconds"
    ^ 31557600 * RelativeOrbitalPeriod at: aPlanet
```

jaccarmac: Oh yeah I guess moving the multiplication out would have made sense. Re: the class methods I could have sworn @macta made comments on my PR about that but looking back I misremembered. My bad!

macta:
Guys - sorry this is turning out to be a slightly messy PR - a problem with iceberg (merging a branch) and trying to use the config generator (to simplify it going forward - but had a side effect of pulling in some other exercise changes by mistake).

Anyway - bear with me, it will be much cleaner going forward.

dev/src/Exercism/SpaceAge.class.st
bencoman: >  Re: the class methods @macta made comments on my PR about that 

That is still good advice - I am guessing it was in relation to old code... https://github.com/jaccarmac/pharo-smalltalk/blob/old-space-age/exercises/space-age/SpaceAgeTest.class.st#L20-L22
``` 
SpaceAgeTest >> testAgeOnEarth [
	self assert: (SpaceAge on: Earth at: 1000000000) closeEnoughTo: 31.69
```
where operation `on:at:` would be implemented on the class side.

Notice with my code that the class method is only for class initialization.
Operational method `orbitalPeriodOn:` is on the instance-side.

One rule of thumb (not unanimously subscribed to) is that if a method doesn't reference any instance state, put it on the class side. But from my scattered review of the exercises, many(/all?) are all so simplistic they don't store any state - so while they "could" be implemented purely on the class-side by the rule, it would provide poor practice for normal stateful programming patterns.

Note Smalltalk has three object/class related variable scopes...
- **instance-variables** - only accessible from instance-side
- **class-variables** - accessible from both instance-side and class-side e.g. RelativeOrbitalPeriod 
- **class-instance-variables** - only accessible from class-side.  Consider that classes are objects too, so they have their own instance variables (important for the system infrastructure rarely used by end-user-developers. For example, see at `methodDict` after doing `Rectangle inspect`)

macta:
@bencoman @jaccarmac - in an effort to close this one down (hope you don't mind @jaccarmac), I did a quick attempt using a class variable (as per Ben's comment). I like the idea of showing the class variable to help educate users on this front. While Ben indicated you can reference the class var directly, it can be error prone as you need to rely on the class #initialize method (which is only called when the class is initialized on load - and not when you modify it) - so this can be very confusing for users if you make a mistake in your table and then don't understand why modifying those values doesn't take hold. So perhaps the proposed is the best compromise? I also left of an instance side getter as we only have a single call (if more used, I would have added an instance method that call the class method).

What do you guys think? Go with this, and lets move on to the next exercise?

bencoman:
> class is initialized on load - and not when you modify it) - so this can be very confusing for users 

Good point!  My immediate reaction was to agree fully with you, but another way to deal with it is a method comment: "Evaluate (SpaceAge initialize) when modifying this method".  Its a good concept to teach, but I'm happy if its left to a later exercise or revised later.  Don't let me drag this out.

bencoman:
Looked more closely at the last commit, it looks like you are doing lazy initialization on the class side.  This will suffer the same problem - the class variable isn't set to nil when the data definition is updated.  The best alternative might be lazy initialization on the instance side i.e. `orbitalPeriods` as an instance variable.

jaccarmac:
No worries on the PR, there are plenty more exercises for me to work on.

Re: class variable initialization: It's indeed an interesting (and arguably important) part of Smalltalk to understand. I was fine with eager initialization but the fact that changes don't update objects even when the class variable is `nil` is really tricky and as a beginner I'm still not sure how to choose the right combination of `initialize*` methods, lazy initializers, etc.

bencoman:
On Mon, 4 Mar 2019 at 13:58, Jacob MacDonald <notifications@github.com>
wrote:

> Re: class variable initialization: It's indeed an interesting (and
> arguably important) part of Smalltalk to understand. I was fine with eager
> initialization but the fact that changes don't update objects even when the
> class variable is nil is really tricky and as a beginner I'm still not
> sure how to choose the right combination of initialize* methods, lazy
> initializers, etc.
>
The argument in most other languages regarding lazy-initialization is
usually couched in terms of efficiency "don't calculate or allocate space
you don't need until you need it"

The main purpose of lazy-initialization in Smalltalk is different because
its a live system - i.e. it never stops (when you close down, its
suspended, not stopped).
Ignoring the class-side of things for the moment, consider you have a
bundle of objects that are constantly in use (ie, the ones used for the IDE
itself, like text boxes, sliders, etc)
and you want to add functionality requiring an extra instance variable.
The moment you add that to the class definition all existing objects
suddenly gain that variable with a nil value.  To avoid scattering #ifNil:
throughout your code, its better to do it it just one place - in the
accessor.
http://onsmalltalk.com/lazy-initialization-smalltalk-style

The live-ness of Smalltalk is why lazy-initialization is more common than
other languages.  It allows more flexibility to grow your objects
on-the-fly.

The definition I read from Java regarding "eager initialization"  is that
it "creates an instance when the class is loaded"
which actually is what happens via the class-side `initialize*` method,
except with Java editing/compiling/running the method
presumes the program exited and the class is loaded again in a fresh
runtime, whereas with Smalltalk the program doesn't exit.

Because a Smalltalk program never stops, you sometimes need to reset the
state of class variables.
One way to do that is manually calling ``MyClass initialize``.
My personal preference is lazy initialization with ``MyClass reset``
resetting class variables to nil - no real justification for that, just the
way the penny fell.

macta:
@bencoman I see we both thought to ask on the Pharo mailing list - which is a nice offshoot of this exercism exploration as I think it will help make the product better. We will see what comes back today - and unless something more compelling comes out - I think the lazy initialisation - possibly with the example >>> designator that does exactly what you suggest, by making the variable nil, is the best all rounder with questions left up to the student to ask more.

macta:
@jaccarmac I'd love you to crack another exercise as you did a great job on this one

jaccarmac:
Sounds good, I'll wait for that patch to land. Also eager to crack some new exercises.

macta:
I'm going to push this one out - and we can always adjust it later if need be. There are beta users that are running out of exercises...

PR #176, opened 2019-03-05
Improved exercism exercise model

macta:
This improves how meta information for exercises is specified (not via class methods on the exercise test, apart from version and uuid which are generated from the spec)


macta:
@bencoman @samWson - might be worth a peak to see how exercises can now be specified (its a bit more compact). I'm going to push this through so I can try another exercise and regnerate the old ones

PR #178, opened 2019-03-05
Get the wordified name from the exercise object

macta:


macta:
This gets everything up to date and is a good starting point for writing new exercises now

PR #182, opened 2019-03-06
Word count exercise

macta:
This solution has an alternative (categorised) - we might want to flag these in the future when we get mentor tools running

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,7 +2,7 @@
 	"track_id" : "pharo",
 	"language" : "Pharo",
 	"active" : false,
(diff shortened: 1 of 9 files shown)


macta:
I'm going to adopt an overnight principle - if others can review overnight (their day), then great, otherwise I'll keep pushing these and can post correct if there are comments. 

As we are still in alpha, I think the concept of review is a bit fluid at the moment anyway. However I like the dicsipline of trying to do these in little chunks (and they are getting smaller and more efficient now)

bencoman:
Sounds good.  I should be able to get back into it middle of April.

PR #207, opened 2019-03-18
Grade school exercise

macta:
@bencoman @samWson - this is an interesting exercise, its the first one that hints at objects. But I've deliberately not created a Student object as it didn't warrant it for a name, grade pair.


exercises/grade-school/.meta/solution/GradeSchool.class.st
```diff
@@ -0,0 +1,58 @@
...
+]
+
+{ #category : #exercism }
+GradeSchool >> addStudents: aCollection [
+	^ self formatResult: (self addAllStudents: aCollection)
+]
+
+{ #category : #exercism }
+GradeSchool >> desiredGrade: gradeInteger [
+	"Anser the Collection of students filtered by grade, alphabetically"
```
samWson: Typo here: `Anser` instead of `Answer`.

samWson:
> this is an interesting exercise, its the first one that hints at objects

Kind of a bit funny when everything is already an object. Looks good to me. I had a look at the Ruby version for comparison. They are using a Hash internally for the students but the actual student could be anything. Looks like your version is storing them as an association which is a good enough object for the lightweight requirements of the exercise.

exercises/grade-school/.meta/solution/GradeSchool.class.st
samWson: Looks like you have already merged. This isn't critical and can be fixed later.

macta:
Thanks @samWson - I've corrected that. Hopefully we can crank out a few more quite rapidly and get us closer to launch position. I've got a few active users on Discord that are enjoying it, and giving some helpful feedback (so far nothing tragic).

PR #212, opened 2019-03-20
Provide a simple mentee solution browser (shadow browser, so doens't …

macta:
…load solution) so you can more easily review a user solution.

@samWson @bencoman This one is pretty cool - it took more code to rearrange the menus than actually do it - although I'd really like to use the proper Iceberg diff viewer as at the moment this is the older Monticello shadow browser (still better than nothing).

image


bencoman:
Great work Tim (I'm feeling such a laggard... ;)
How do you select a particular mentee?

PR #223, opened 2019-03-29
Regenerated the remaining tests to pickup the test numbering improvement

macta:


bencoman:
Where do we advise students that numbering tests like this is not usual?
Should each exercise have a `test00_note` method that advises something like...
"Its not usual to number tests like this.  Such test numbering is an artifact of providing an order or engagement for the Exercism problem"

PR #225, opened 2019-03-29
Improve the testing documentation

macta:
As per Ben's concern in #223 - we hadn't previously documented why we have testXX_ methods. 

@bencoman see if this helps a bit

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -4,7 +4,7 @@ At the most basic level, Exercism is all about the tests and
 [testing](https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md).
 
 In Pharo, you can run tests for a single exercise by clicking on the coloured test orb next to any TestCase class or method.
-The orb will then be colored according to the result of the last test run: 
+The orb will then be coloured according to the result of the last test run: 
 - green for pass 
 - yellow for an assertion failure 
 - red for a runtime error/exception.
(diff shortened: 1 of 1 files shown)


docs/TESTS.md
```diff
@@ -3,13 +3,27 @@
...
+In Pharo, you can run tests for a single exercise by clicking on the coloured test orb next to any TestCase class or method.
+The orb will then be coloured according to the result of the last test run: 
+- green for pass 
+- yellow for an assertion failure 
+- red for a runtime error/exception.
 
-Larger groups of tests can also be run using the Test Runner tool from the world menu or typing `<meta> + OU`.
+The tests in Exercism exercises have been purposely numbered to give a defined order of execution (test01_verifySomeProperty, test02_verifyAnotherProperty etc.). When working on an exercise it is recommended that you click on the orb of the first test, understand the failure and what is required to make it pass and then add the code to your solution to make the test pass. In Pharo it is quite normal to make these changes in the debugger where you have access to both the code editor as well as a view of all the variables and parameters that can help you understand the problem. 
+
+_NOTE: It is not a regular practice to label tests with an ordering prefix, and you should NOT do this when writing tests for your own code._
```
bencoman:
```suggestion
_NOTE: It is not a regular practice to label tests with an ordering prefix, and you should NOT do this when writing tests for your own projects._
```

docs/TESTS.md
```diff
@@ -3,13 +3,27 @@
...
+
+Pharo is also happy to run with code that is broken, and the debugger will simply re-open when code is encountered that is in error (either a syntax error, bad value or even a missing class or method). This technique was a primary influencer for test driven development, and you can get a taste of this approach when you run the first test of any exercise. The debugger will immediately report that your solution class is not found (as you haven't tried to solve it yet). Conveniently, their is a "Create" button that helps you add the missing class and it then continues running the test until it either finishes, or another error is encountered. With your first test, you normally then encounter a second error for a missing method in your new class. Again, the "Create" button in the debugger lets you define it, and continue running. At this point, you can then click further down the stack trace and review the requirements of the failing test and modify your freshly created method to make that test pass.
```
bencoman: > Pharo is also happy to **run** with code that is broken
Pharo is also happy to **work** with code that is broken
```suggestion
Pharo is also happy to work with code that is broken, and the debugger will simply re-open when code is encountered that is in error (either a syntax error, bad value or even a missing class or method). This technique was a primary influencer for test driven development, and you can get a taste of this approach when you run the first test of any exercise. The debugger will immediately report that your solution class is not found (as you haven't tried to solve it yet). Conveniently, their is a "Create" button that helps you add the missing class and it then continues running the test until it either finishes, or another error is encountered. With your first test, you normally then encounter a second error for a missing method in your new class. Again, the "Create" button in the debugger lets you define it, and continue running. At this point, you can then click further down the stack trace and review the requirements of the failing test and modify your freshly created method to make that test pass. 
```

bencoman:
Looks good - only added some minor wordsmithing...

docs/TESTS.md
macta: @bencoman I realised I cocked this up, as I pressed merge without accepting your change... I've corrected that now (didn't bother with a branch). I also added some headings, as when you see it on the live site (https://exercism.io/tracks/pharo-smalltalk/tests) it was a bit wordy.

Happy to go with any other changes too - but its definitely better from your prompting.

PR #241, opened 2019-04-01
235 add a solution for the clock exercise

macta:
Integrate @raydeejay solution (his pr is locked)


macta:
replace by a new branch

PR #254, opened 2019-04-03
Flip the active switch

macta:
Change the config template to active

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,8 +1,8 @@
 {
 	"track_id" : "pharo",
 	"language" : "Pharo",
-	"active" : false,
-	"blurb" : "Pharo is a fun and immersive smalltalk programming experience. Featuring a clean object-oriented language and a powerful user extendable environment, it focuses on simplicity and immediate feedback (think IDE and OS rolled into one). It's elegence and productivity give it wide reach into banking, education, web development and even robotics.",
+	"active" : true,
+	"blurb" : "Pharo is a fun and immersive Smalltalk programming experience. Featuring a clean object-oriented language and a powerful user extendable environment, it focuses on simplicity and immediate feedback (think IDE and OS rolled into one). It's elegence and productivity give it wide reach into banking, education, web development and even robotics.",
 	"solution_pattern" : "\\.meta/solution/\\w+\\.class\\.st",
 	"test_pattern" : "\\w+Test\\.class\\.st",
 	"exercises" : [
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "track_id" : "pharo-smalltalk",
   "language" : "Pharo",
-  "active" : true,
+  "active" : false,
   "blurb" : "Pharo is a fun and immersive Smalltalk programming experience. Featuring a clean object-oriented language and a powerful user extendable environment, it focuses on simplicity and immediate feedback (think IDE and OS rolled into one). It's elegence and productivity give it wide reach into banking, education, web development and even robotics.",
   "solution_pattern" : "\\.meta/solution/\\w+\\.class\\.st",
   "test_pattern" : "\\w+Test\\.class\\.st",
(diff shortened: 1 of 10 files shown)


macta:
@iHiD @kytrinyx - gulp

@ErikSchierboom, @iHiD said you might have a cursory look at this.

@samWson , @bencoman - this might be it guys!

bencoman:
Super cool.  Thanks for your great work getting it to this point.
btw, on the weekend I'm travelling China for 10 days.
After that I'll run myself through it as a student, then jump in as a
mentor.

macta:
> Super cool. Thanks for your great work getting it to this point. btw, on the weekend I'm travelling China for 10 days. After that I'll run myself through it as a student, then jump in as a mentor.

Enjoy China (one on my hitlist) - we will have a decent selection of exercises, and all mentoring help appreciated (I've got raydeejay warming up on this too - he's been pretty active helping answer questions on Discord)

config.json
```diff
@@ -1,370 +1,370 @@
...
+      "slug" : "two-fer",
+      "uuid" : "9806fcc0-8505-4012-bd64-3f7468014df5",
+      "core" : true,
+      "auto_approve" : false,
+      "unlocked_by" : null,
+      "difficulty" : 1,
+      "topics" : [
+        "strings",
+        "text-formatting",
+        "refactoring"
```
ErikSchierboom: Is the `two-fer` solution really about refactoring?

ErikSchierboom:
One tiny thing I noticed: I checked out the solution on Windows, and then used WSL (built-in Linux) to run `./bin/fetch-configlet`. This did not work, as that file was checked out with CRLF line endings. So perhaps an addition to the `.gitattributes` could be made?

All in all, great work in getting this track (almost) live!

config.json
macta: Actually I have found that it is a great  opportunity to bring up refactoring as many people don’t spot that the default  case can call the specific one with “me”. So I tagged it that way

macta: I hadn’t spotted the binary deprecation - one to add to our generator to check for later deprecations (had forgotten that). (I recall binary being quite good - it’s a nice example of collecting results then summing them) 

The idea was to keep filling in the blank unlocking ones, but I’ll check out your ideas as they sound good.

You are right about bracket push too - it’s easy when you know how, but the concept Is trickier. I need to finish more harder ones so it gets a good place

macta:
Tell me more about “checked out the solution”? I ask because Pharo is a bit unique in that the installation instructions guide you to download a small cross platform environment that has menus to fetch exercises via http (and the exercism endpoints).

So yes - it’s possible the line endings are wrong. I don’t think the Pharo compiler cares, I will see if it’s a flag when we write our code - but maybe I can use .gitattribtues - I’m not super familiar with that (we have one  for other files, but maybe I can set it for .st files too - is it just “ *.st text”?)

ErikSchierboom:
> Tell me more about “checked out the solution”? 

I mean I just cloned the repository to my file system. But it not something important, as people using WSL will know how to deal with this. It does not actually have anything to do with the exercises, it is the `./bin/fetch-configlet` file in the repository.

macta:
@ErikSchierboom so you mean the bash file `fetch-configlet` has the wrong line endings? Do I just name that file in the .gitattributes file?

ErikSchierboom:
> Do I just name that file in the .gitattributes file?

I think you do (I'm not that familiar with `.gitattributes`). 

>  And possibly I should add *.st there too?

This depends on the fact if editors can deal with different types of line endings. The bash file won't work unless the line endings are LF, but usually other files work fine.

iHiD:
@macta Would you mind changing the config.json to have `active: false` - merging this and checking it's all good, then opening a PR with only that one change in it please :)

macta:
Separating fixing live feedback from go live switch

PR #266, opened 2019-04-06
Update ReverseStringTest

chicoary:
{ #category : #extra }
ReverseStringTest >> testMisuseOfSelector [
	self verifySolution: reverseStringCalculator class avoidsSelector: #reverse.
	self verifySolution: reverseStringCalculator class avoidsSelector: #reversed
]


macta:
Nice spot - thanks

PR #271, opened 2019-04-10
Resistor color duo exercise

macta:


macta:
This is pr is to fix https://github.com/exercism/problem-specifications/issues/1495 which renames this exercise

Issue #263, opened 2019-04-05
Add a solution for the Darts problem


raydeejay:
Working on it

PR #278, opened 2019-04-14
263 add a solution for the darts problem

macta:
Another contribution from @raydeejay 
Improves the generated tests for darts


macta:
fixes #263

Issue #264, opened 2019-04-06
Add a solution for the Collatz Conjecture exercise


raydeejay:
Another easy one :-)

PR #279, opened 2019-04-14
264 add a solution for the collatz conjecture exercise

macta:
Another @raydeejay exercise


macta:
Changed the name of the result method to make it more smalltalk like

macta:
Closes #264

PR #282, opened 2019-04-17
Add base implementation of Trinary exercise.

carlotxra:
Add base implementation of Trinary exercise.
Rename generated test methods

### DIFF
--- a/dev/src/Exercism/Trinary.class.st
+++ b/dev/src/Exercism/Trinary.class.st
@@ -1,0 +1,28 @@
+Class {
+	#name : #Trinary,
+	#superclass : #Object,
+	#category : #'Exercism-Trinary'
+}
+
+{ #category : #'private validating' }
+Trinary >> isValidTrinaryValue: anIntegerOrString [
+	^ '0123' includesAll: anIntegerOrString asString
+]
+
+{ #category : #private }
+Trinary >> parseTrinaryValue: anIntegerOrString [
+	| trinaryValue |
+	trinaryValue := 0.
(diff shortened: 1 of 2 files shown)


macta:
This is cool - I'm on the road, but will also generate the exercism meta files tomorrow (ish) and we can push this one out. Appreciate the help

carlotxra:
No problem. Still trying to understand the iceberg integration. Also see that travis broke for some reason.
Let me know if I missed something or if you want me to change something.

macta:
Not sure why Travis broke - seems like a time-out just running the first test .

macta:
Interesting - I see that you must have edited the test method names? They are a bit verbose - and it actually suggests an improvement to the generator, to only use the description name if there are multiple sub cases.

So I should fix the generator as if we need to regenerate, the names will revert back ....

macta:
You've also highlighted another bug - as it turns out that Exercism has now deprecated Trinary - so I'll come back to this one and figure out if we can make it a bonus exercise.

bencoman:
@macta what are your thoughts now on this one? (to clear PRs)

macta:
I believe i fixed the generator based on what @carlotxra had done (its not identical but very close) - I'd be up for regenerating, apply the solution and making this a bonus exercise anyway - having lots to play with is good for folks like @RoelofWobben

RoelofWobben:
why ?

bencoman:
go ahead. regen would be better.

Bajger:
Exercise deprecated, should be implemented by All Your Base exercise.

Issue #255, opened 2019-04-04
Some error http conditions don't give exeception, just a web page with forbidden


macta:
With some proxies - you don't get an http error and so a get results in a page like below.  Might need to detect this kind of string and generate an applicaiton execption:

```
<html>
	<head>
		<title>Forbidden access</title>
	</head>
	<body style="text-align:center">
		<div style="width:80%; border:1px solid #64696C; color:#64696C; font-weight:bold; text-align:center; padding:20px; margin:auto; margin-top:20px">
			<p id="small_p" style="font-size:small;">Forbidden access</p>
			<p id="large_p" style="font-size:large;">Your current web browser cannot access this page.</p>
		</div>
	</body>
</html>
```

macta:
This might not happen if we better config the znClient. Sven alerted me to using : 
```^ ZnClient new
		systemPolicy;
```

This also means we need to redo the exception handling (it actually simplifies it)

PR #301, opened 2019-04-26
Document exercism API

### LINKED ISSUE #281: Document our use of Exercism REST API
This occurred to me as I was thinking about working on #245. The Exercism REST API, which we make use of, isn't documented. We have effectively discovered it ourselves through our work on this repo and properly documenting how we use it would help with future maintenance.

Documentation would also have the side benefit of being useful to the Exercism maintainers and any other language track maintainers who wished to use the API as well.

samWson:
Fixes #281.

This is work in progress for the moment. 

I've started with fetching exercises for now. Documenting the details of a `200` response is a bit difficult for me. I'm not that familiar with REST APIs and trying to get an idea of what it would look like from reading our own code is hard too.

Checking the exercism website source code has been very useful for getting the range of possible responses.

More to follow in this PR. The details of a `200` response need to be written and submitting exercises needs to be documented.

### DIFF
--- a/docs/exercism-api.md
+++ b/docs/exercism-api.md
@@ -10,12 +10,15 @@ All Endpoints require headers on the request to be set to the following values:
 
 ```
 Authorization: Bearer <token>
-Content-Type: application/json
 ```
 
 The value `<token>` is the users CLI authentication token available at `https://exercism.io/my/settings`.
 
 ## Fetching Exercises
+
+```
+Content-Type: application/json
+```
 
 `GET api.exercism.io/v1/solutions/latest?exercise_id=<name>&track_id<language> HTTP/1.1`
 
@@ -31,7 +34,7 @@ A sucessful request will respond with the exercise.
 
 #### 400
 
-The language track name is ambiguous.
+The language track name is ambiguous, or the submitted exercise files have not changed.
 
 #### 403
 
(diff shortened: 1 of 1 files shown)


macta:
This is great - particularly if its an itch you want to scratch to learn more.

Kind of related is: https://github.com/exercism/pharo-smalltalk/issues/255 as we aren't doing the http stuff very cleanly. There is also the question of why exercism doesn't use 302 vs 400 (as mentioned in: https://github.com/exercism/pharo-smalltalk/issues/281)

samWson:
Added some more info. Still have the fine details of what the download and submit payloads look like to work out.

samWson:
@macta @bencoman I'm thinking I might merge this PR as it is. I've managed to document quite a bit. The only thing missing is detail on submitting exercises which I can revisit later. Better to have this documentation as it is than not at all.

Have a look and let me know what you think.

macta:
Sure - I also just answered a question from Ben on generation - might be worth having a read and possibly folding it in to this?

bencoman:
@samWson I read the final in master:/docs.  Very nice.  I particularly like you sussed out the info on the return codes - they were unclear to me.

Only minor suggestion, `<name>` might be better as `<slug>`, for how Exercism refers to the kebab-case version of the exercise name.

samWson:
@bencoman thanks for the feedback. I'll make a fix. I'm taking the opportunity now to try the exercises myself. It's why I wanted this language track in the first place.

Issue #248, opened 2019-04-02
Add a solution for the ISBN Verifier exercise


raydeejay:
As usual, adding a solution makes the exercise acceptable.

PR #304, opened 2019-04-26
248 add a solution for the isbn verifier exercise

macta:
Another submission from @raydeejay


macta:
I'm merging to get the backlog of exercises out of the way so we can do some fixing

macta:
Fixes #248

Issue #302, opened 2019-04-26
Move Exercise tags over to full Packages


macta:
Our use of pharo tags (e.g. Exericism-HelloWorld) was a nice way to organise exercises in the image (so they can collapse and also provide a top level menu item) - however the direction of pharo is to keep tags a light weight categorisation and instead rely on a Project for organisation. 

The current tag usage has some issues we need to address:
1) Extension methods aren't supported in tags and the data is not available to do this
2) Useful tools like the Critique engine also work at the package level

The guidance from upstream pharo is to use Packages, and so we need to port the existing tags to top level packages.

Without projects, its still useful to see all the exercises together (and not mingle in the development/support packages), so we need a useful prefix (later a project might work fine, when that work is complete).

I propose experimenting with Exercism_<name>, which should put all our exercises above the support packages (alternatively we rename ExercismDev, ExercismTest etc to move them away for the exercises).

It shouldn't be much work to do this - the menu detection code will need adjustment, and having an Exercism entry in the new menu bar now addresses the issue with how to load a new exercise if you don't have any loaded. This should also clean up the problem with Welcome dirtying the image (we wouldn't have to move it).

bencoman:
> so we need a useful prefix 
For quite a while I've quietly thought the 'Exercise-'
would be a good way to distinguish exercises from our "Exercism-" dev environment.

macta:
I hadn’t spotted the subtlety of Exercise vs Exercism (but quite like it) ... the issue with the $- separator is it gets interpreted as a tag (unless programmatically applied, which is a pain).

I played with Exercism@ eg Exercism@HelloWorld, which wasn’t too bad, and gave an easy backwards compatibility with menu detection etc.

How do you feel about that? (But could explore ExerciseHelloWorld or Exercise_HelloWorld (or @ too)

bencoman:
Part of this will rearranging what Iceberg perceives as full packages.  Iceberg seems to use disk-folder arrangement and the **packages.st** file to determine what are full packages. I have successfully rearranged this from the command-line - WIP PR incoming shortly for review.

The other part is that our Exercism "downloading" of exercises is independent of Iceberg's arrangement. 
 It only has the [text string of #category:](https://github.com/exercism/pharo-smalltalk/blob/master/dev/src/Exercism/Acronym.class.st#L7) to parse and depends on whether a top-level **Exercism** package exists when each exercise is loaded.  Without pre-creating that, the exercises naturally install as full-packages.  

[...]

PR #311, opened 2019-04-28
Convert exercise tags to packages 

macta:
Currently a spike, and related to similar work @bencoman has done - we are collaborating to move the tags over to full packages to address #302


bencoman:
Its seems you've done more than me in the conversion including supporting methods, so ignore my **fullPackage2** branch and just continue with yours.  Although I notice you went with Exercism@HelloWorld rather than Exercise@HelloWorld.

macta:
@bencoman so I've got it working, with tests passing (although the coverage is still not quite idea). It will load both as a default baseline (and you can then load the dev baseline - what a pain to implement with so many packages - see the BaselineOf).

This will work with both old and new formats - so we can direct existing users to follow the upgrade steps we have already written (their existing exercises will stay and can still be submitted), but new ones will adopt the new package format.

Sorry - I misunderstood on the Prefix@Name, I thought you meant Exercism vs Exercise - but it should be quite easy to change it (as its a script). Do you want to do that?

macta:
@bencoman thanks for chipping in on this, it makes it much easier having more minds on this. Let me know if you think exercise@helloworld resonates better - its a pretty simple fix, and then we can do a more testing and see how close we are to pushing something out.

bencoman:
On Tue, 30 Apr 2019 at 01:36, Tim Mackinnon <notifications@github.com>
wrote:

> @bencoman <https://github.com/bencoman> thanks for chipping in on this,
> it makes it much easier having more minds on this. Let me know if you think
> exercise@helloworld resonates better
>
With capital 'E', yes I think it is better.

> - its a pretty simple fix, and then we can do a more testing
>
and see how close we are to pushing something out.
>
I've been wondering how we go about testing without impacting existing
users.
Perhaps temporarily generate a few like /exercise/zzz-experimental-anagram
so we ourselves can check it works as expected before exposing users to it.

cheers -ben

RoelofWobben:
and im willing to helping with testing but I have to know which steps I have to takeso I get the new challenges

macta:
@bencoman @RoelofWobben - I think this one is ready to merge into master, and along with Die (a custom exercise) we can prove it works with the upgrade instructions of new and old exercises

RoelofWobben:
as far as I can test the upgrade and install instuctions still work.

macta:
I'm going to merge this to get @RoelofWobben to help with testing as other pieces are starting to require it (like the TwoFer critics, and a few others neading extension methods)

PR #312, opened 2019-04-29
Add submit instructions (with picture) to make it clearer how to submit

macta:
@samWson this was some feedback from Stef who was stuck on how to submit


samWson:
Looks good. I'm guessing there is no real way to see if the links load and the images look good other than just merging and seeing the result?

macta:
Actually you can check the images - follow the commit link and naviagate down to the image (you just have to specify in the text, the final live url).

e.g. https://github.com/exercism/pharo-smalltalk/blob/submit-exercise-instructions/docs/images/SubmitExercise.png

Issue #286, opened 2019-04-19
spelling "langauge"


bencoman:
To get back into this, I'm running through the exercises as a student.  Noticed a spelling error in "langauge" but it looks like the same text duplicated in four locations.... https://github.com/exercism/pharo-smalltalk/search?q=langauge

Have they been generated from somewhere?

macta:
You are correct its generated - the Hint text comes from the class comment of each exercise. The "master" version is the class comment of the exercise - the 2 duplicates come from how exercism processes things with its "configlet" cli tool (hence readme and hints.md), the other 2 are: the normal pharo tonel src, and the final is where we drop the file so exercism knows how to serve it up  (as it doesn't know about how we save src files).

In retrospect - we might be able to clean up (or exclude hints.md) - I don't think exercism would miss it after generating things. We might also be able to symlink the exercism/exercise files back to our superset src files (but not sure how github handles links, and whether it would bugger windows users trying to contribute)?

This also reminds me - I run the generation of these files locally, but I wonder if our ci could do it somehow (but then our ci would need to create another verision for Exercism the platform, so it gets a bit complicated?)

macta:
Lost in all that - to fix: correct the class comment in HelloWordTest. Ensure you have your dev environment confgured per readme, and then run "Exercism|ReGenerate meta data" menu and checkin.

macta:
I checked on the Exercism slack - and I think the platform uses a post commit hook to retrieve the files it wants to then do its track staging... this said, we once had a separate repo (for a slightly different reason) - but if we did move all of the development over we could have a ci job to generate and push the final result (OR maybe we could develop on a different branch - and have a job that pushes the final files to master for exercism to process - thinking about it, that might be tidier?)

PR #314, opened 2019-04-30
spelling fix

gunnaraas:
langauge -> language


macta:
See also https://github.com/exercism/pharo-smalltalk/issues/286

macta:
We got another submission that did this one in the actual src file - so will remove this. Thanks for bringing it to our attention

PR #327, opened 2019-05-10
Update Diehandler

RoelofWobben:
I made this pr because code critiz does not like that the method has the name D20.

### DIFF
--- a/exercises/die/DieHandleTest.class.st
+++ b/exercises/die/DieHandleTest.class.st
@@ -37,14 +37,14 @@ DieHandleTest >> testRoll [
 
 { #category : #tests }
 DieHandleTest >> testSimpleHandle [
-	self assert: 2 D20 diceNumber equals: 2.
+	self assert: 2 d20 diceNumber equals: 2.
 ]
 
 { #category : #tests }
 DieHandleTest >> testSumming [
 
 	| handle |
-	handle := 2 D20 + 3 D10.
+	handle := 2 d20 + 3 d10.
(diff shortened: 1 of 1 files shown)


bencoman:
I suggested this, so thanks @RoelofWobben for doing the PR. 
@macta, do I correctly understand this exercise is not in the problem-specifications, so will not be overriden by later test generation?  I can't see this cahnge causing a problem, but since you only added the `die` exercise recently, so I'll leave you the merge to you to avoid conflict with anything else your doing with it.

macta:
It’s a good change - although this exercise solution needs extension methods that we need to fix next (but this helps too).

Regeneration for custom exercises is fine - I updated the readme.md with the basics

RoelofWobben:
yep, when the solution contains the extensions methods I will make a PR to make that work

macta:
Argggg - I forgot, you can't fix this here - this is the generated version and so I've just generated over top of it. You have to fix it in the src/ directory.

@bencoman unfortunately Exercism don't have a provision for running configlet as a post process hook on their end, so we have to do it (currently manually) but maybe we could have a seperate repo for all our stuff, which then commits the final output to the exercism repo automatically (but then you could still commoit into that repo by mistake too)

macta:
@RoelofWobben fancy doing it again - but in the proper tree (otherwise I can do it for you)

RoelofWobben:
Np

so to be sure. I have to change this one : https://github.com/RoelofWobben/pharo-smalltalk/blob/master/exercises/die/DieHandleTest.class.st

PR #329, opened 2019-05-11
custom-QA-rule

bencoman:
Example custom Quality Assistant rule for reuse in TwoFer

### DIFF
--- a/dev/src/Exercism/ExercismHintRule.class.st
+++ b/dev/src/Exercism/ExercismHintRule.class.st
@@ -1,0 +1,16 @@
+Class {
+	#name : #ExercismHintRule,
+	#superclass : #ReAbstractRule,
+	#category : #'Exercism-TwoFer'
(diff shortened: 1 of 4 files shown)


macta:
Nice! Haven't tested it yet - but looks good. Does it present itself as a yellow triangle - or do you have to run the critic?

Armed with some of these - we can plug these into the auto-mentor project (I think we need to work out a docker vm or something, to run an image cmd line to answer the results of our critics). I'd held off as I felt we ought to get our other bits sorted first - however having some of these in place will make it super easy. Thanks for chasing.

macta:
It's really nice @bencoman - we could consider offering to run our exercism tool critics before a submit as well (essentially their auto-mentor locally)

macta:
While we're on a critic role - I'll commit it (we may want to move them to a new ExercismCritic package if we do a few more)

PR #333, opened 2019-05-12
Fixes #184 support-extension-methods-for-submitfetch

### LINKED ISSUE #184: support extension methods for submit/fetch
We don't support extension methods in submit/fetch but there are some exercises like (strain) that are meant with that in mind. Also, I think we should teach people that "monkey patch" is not an evil thing.

Probably should contemplate this after #129 .

I'm thinking we could have a pragma like: <exercism: #ExerciseNameTest> and then we could ensure that methods tagged with this were written out (the issue is that extensions can only live at the package level and not the sub-package/tag level). A bit frustrating as the tags otherwise give a great end-user experience to keep all of the exercises neatly together.

I think the trade-off that a few extension methods sitting in the top level 'Exercism' and the requirment to use the above pragma should be ok?

bencoman:
Fixes #184 - submit extensions methods & improved extension protocol suggestion


dev/src/ExercismTools/ExTonelWriter.class.st
```diff
@@ -14,22 +14,28 @@ Class {
...
+	"Tonel export classes with their methods, mapped from their filename to content streams"
 	(snapshot definitions select: #isClassDefinition)
-		do: [ :classdef |  |tonelStream|
+		do: [ :classDef |  |filename tonelStream|
+			filename := classDef className, '.class.st'.
 			tonelStream := WriteStream on: String new.
-	 		self writeClass: classdef on: tonelStream.
-	  		tonelMap at: classdef className put: tonelStream contents ].
+	 		self writeClass: classDef on: tonelStream.
+	  		tonelMap at: filename put: tonelStream ].
```
bencoman: The caller doesn't have the information to determine which classes are extensions, so we create the ".class.t" filenames here and the ".extension.st" filenames below.

macta:
Quick visual scan looks pretty good - but the test failure needs examination. We should merge in master as well so we can properly test the full changes in an image.

dev/src/ExercismPharo70/DoesNotUnderstandDebugAction.extension.st
```diff
@@ -15,6 +15,25 @@ DoesNotUnderstandDebugAction >> createMissingClassWith: variableNode in: aContex
 	previousContext restart
 ]
 
+{ #category : #'*ExercismPharo70' }
```
macta: @bencoman has this stuff been also submitted as PR in Pharo 8 (e.g. like: https://github.com/pharo-project/pharo/pull/2893). This package is mirroring those PR's so when we hit P8, we remove this one, and start a Pahro80 version for new things we find)

dev/src/ExercismTools/ExTonelWriter.extension.st
```diff
@@ -4,22 +4,28 @@ Extension { #name : #ExTonelWriter }
...
-	
-	"ignore packages (as done in #writeSnapshot:"
-	
-	"now export classes - but to a map"
+		
+	"Tonel export classes with their methods, mapped from their filename to content streams"
 	(snapshot definitions select: #isClassDefinition)
-		do: [ :classdef |  |tonelStream|
+		do: [ :classDef |  |filename tonelStream|
+			filename := classDef className, '.class.st'.
```
macta: Doing this fails tests - I had a look and have a quick fix for you, will try and merge onto your branch (hopefully you ticked allow maintainer edits?).

I was playing with moving this onto ClassDefinition (e.g. classDef exTonelFilename) so a test can access it to (e.g. its effectively a constant).

dev/src/ExercismTools/ExercismExercise.class.st
```diff
@@ -324,23 +324,33 @@ ExercismExercise >> solutionId [
...
-	sources := ExTonelWriter new mappedSnapshot: self snapshot.
+	packageFileMap := ExTonelWriter new mappedSnapshot: self snapshot.
 	
+	"Exclude non-solution classes, i.e. filter out tests"
+	solutionClassNames := self solutionClasses collect: [ :c | c name ].
+	solutionFileNames := packageFileMap keys select: [ :filename | 
+		solutionClassNames includes: (filename findTokens: '.') first ]. 
+
+	"Sort classes before extensions, then sort by class name"
+	solutionFileNames sort: [ :filenameA :filenameB | |tokensA tokensB|
```
macta: why not sort the classes first (then no need to do the token split)?

bencoman: Cool. I just came to a fork in the road... https://www.youtube.com/watch?v=p-mGXLgGqkY
and jumped one way.  Happy to go the other. 
To be specific though, rather than "classes first" it could be considered "purely by filename".

dev/src/ExercismTools/ExTonelWriter.extension.st
bencoman: I liked what I saw that in your commit.

dev/src/ExercismPharo70/DoesNotUnderstandDebugAction.extension.st
bencoman: Good idea.

dev/src/ExercismTools/ExTonelWriter.extension.st
macta: I've tested it as is now the tests are passing, and I think we should move on (we have more important cleanups to do)

dev/src/ExercismTools/ExercismExercise.class.st
macta: I thought I had resolved this bit of conversation? remember that fork well... anyway more other important fixes to do and we can improve this bit later (and also tested it in an image as is), so happy to leave it.

Issue #335, opened 2019-05-13
Setup methods critic error for every challenge


RoelofWobben:
Hello, 

When I look at the two fer challenge tests with crititz I get this two messsages. 
 
1) provide a call to super setup as first message in the Setup Method

I can make a PR for this one. 

2) inconstent method classification of the Setup

the setup method is now in the Setup protocol. 

and according to the text this is not right 

All methods should be put into a protocol (method category) that is equivalent to the one of the superclass, which is a Smalltalk style convention.

which is correct

On the challenges setup is on the setup protocol but on a superclass named TestCase it is on the running  protocol

RoelofWobben:
also the class method exercise has for all exercises not a good protocol

macta:
Fixed in the generator by https://github.com/exercism/pharo-smalltalk/pull/347 but still needs some method rewrite magic as the generator doesn't support regeneration yet

PR #349, opened 2019-05-15
Fix tests not calling super setUp, and also having wrong category. 

macta:
Also moved exTonelFile extension method to correct package


macta:
Fixes #335

PR #358, opened 2019-05-16
fix typo on hello world challenge

RoelofWobben:


macta:
Your getting the hang of it... thanks

PR #361, opened 2019-05-17
Experiment to share code on dpaste

macta:
Something for @RoelofWobben to try when asking for help.
Doesn't quite work as ";" seems to cause the source to get cut off?


macta:
fixes #363

macta:
This solution is implemented in dPaste (but there are examples with stfx and paste.ee - however dPaste seems to best fit the bill)

PR #362, opened 2019-05-18
Better legacy exercise support

macta:
Issue reported by @RoelofWobben - can't submit a legacy exercise due to removal of Exercism parent package


RoelofWobben:
tested on a new and old package and found no problems

PR #366, opened 2019-05-23
First pass at mine-sweeper

macta:


bencoman:
I've always thought `#eightNeighbors` was a cute utility method.
e.g. Try... `(5@5) eightNeighbors inspect`

While this could be an opportunity to introduce Array2D (although I've never used it), 
if a bit of domain knowledge is pushed into Array it could be solved quite elegantly like this...
```
Array>>mineAt: aPoint 
	(aPoint y < 1) | (aPoint y > self size) ifTrue: [ ^0 ].
	(aPoint x < 1) | (aPoint x > self first size) ifTrue: [ ^0 ].
	^ ((self at: aPoint y) at: aPoint x) = $*
		ifTrue: [ 1 ]
		ifFalse: [ 0 ]
```

In Playground...
```
minefield := #(
   '***'
   '*  '
   '***' ).

result := minefield withIndexCollect: [ :row :y |
    row withIndexCollect: [ :cell :x |
        cell = $* 
            ifTrue: [ $* ]
            ifFalse:  [ 
                count := ((x@y) eightNeighbors collect: [ :p | minefield mineAt: p ]) sum.
                count = 0 
                    ifTrue: [ Character space ]
                    ifFalse: [ count asCharacterDigit ]
                ]
        ]
    ]
```
I found #asCharacterDigit using Tools > Finder > Examples...  3 . $3
Out hints might direct students to do the same, 
and also inform about `Character space`, maybe with example... `('a b' at: 2) = Character space` ==> true

bencoman:
Or pushing a little bit further...
```
Integer>>asMinesweepDigit
    ^ self = 0
            ifTrue: [ Character space ]
            ifFalse: [ Character digitValue: self]
```
facilitates...
```
result := minefield withIndexCollect: [ :row :y |
    row withIndexCollect: [ :cell :x |
        cell = $*  
            ifTrue: [ $* ]
            ifFalse:  [ ((x@y) eightNeighbors sum: [ :p | minefield mineAt: p ]) asMinesweepDigit ]
        ]
    ]
```

macta:
@bencoman nice comments and good feedback (its interesting how such a simple example can show you so many interesting things - I'd never heard of eightNeighbours  - presumable something for doing some bitBlt operation). 

Aside: id actually intended to this one on the train today, but got sucked into it last night - then revisited based on your feedback.

The only real difference now is whether Array2D is interesting, or whether just working over a standard Array is clearer (not sure - probably depends on how your brain works - although I would say, Array2D seems a bit half finished - e.g. surprised you can can't index on a point, or directly iterate over rows/columns).

macta:
Actually, on reflection - Ben is right - Array2D is not needed (although it's not a bad approach), however sticking with an Array makes this very simple

PR #367, opened 2019-05-25
First pass at matrix exercise

macta:
This one is a precursor to minesweeper


macta:
Its not clear if this exercise should lead you to create a cheap Matrix class that holds contents and allows multiple operations on it, or if the point is just to process some string data in row and column format. (see my comment at : https://github.com/exercism/problem-specifications/issues/1496) - I would argue the problem definition points you to the latter which is how the generator has created it. However other languages like C# and Python have created an object to hold the data.

PR #368, opened 2019-05-27
Initial Flatten Array exercise

macta:


bencoman:
> Without using the built-in flattening method

A passing thought... I wonder if we need to reevaluate our policy of not-using built-in library functions.
Our aim is more to induct students into Pharo than artificially jump hoops just to teach problem solving.  Its not our fault our rich class library makes some of these problem trivial.  ;)
Of course we wouldn't want to do that for all the trivial problems, since those are simple exercises for newcomers to work through.

For some middle ground, I see you use streams and looking through the Collection library I notice #flattenOn: that I've never used before.  Maybe students are asked to subclass WriteStream - because that is the power of Pharo and builds a sense of mastery in students.
Untested concept...
```
NilFilteredWriteStream >> nextPut: anObject
    anObject ifNotNil: [ super nextPut: anObject ]

FlattenArray >> flattenArray: aCollection [ 
      |stream|
      stream := NilFilteredWriteStream on: OrderedCollection new.
      aCollection flattenOn: stream.
      ^ stream contents asArray
```
The question then is how to coach students to do it this way.  
Can we perhaps seed the exercise downloaded students with `FlattenArray >> flattenArray:` as above,
and the task for students then devolves into a more Pharo-like task of subclassing NilFilteredWriteStream properly (for which there can be Exercism QA hints) and overriding #nextPut:

bencoman:
btw rereading, I'd make that **MyNilFilteredWriteStream**
which I think makes the missing class less confusing.

P.S. Editing comments is broken from this browser at work.

macta:
I like the idea of suggesting a stream subclass in the hint , and removing the invalid selector check.

macta:
I actually thought similar, but as it turns out - flattenOn: won’t work as the tests expect you to ignore nil’s.

I looked at python and ruby and C# when I did it, and 2 of the 3, followed the Exercism ethos of not using the built in library functions (I think ruby ignored it).

The earlier exercises where these crop up are actually fun to do by not using the built in functions as they tend to be about teaching you iteration/recursion/streams and Ive been impressed how many different solutions students have come up with when mentoring them.

I think Exercism are eliminating these exercises, which given above, I think is a shame - having a mixture is quite good.

macta:
@bencoman I played with the idea of having multiple solutions in our package - it's reasonably straight forward and so I've added it (although it relies on giving the alternate solution a tag - which is a bit subtle). However, as students never really see this - It's not so bad.

exercises/flatten-array/.meta/hints.md
```diff
@@ -0,0 +1 @@
+Without using the built-in flattening method, you can derive your own, but you will need to differentiate between what is a collection and what is not. Alternatively, if you decide to use the built-in flattening methods, you will need to create a new Stream type to override incorrect behaviour.
```
macta: @bencoman this should give room to gently push users to solve the problem how they see fit.

bencoman: I'd like to be more explicit about the trivial solution being ignored, stating something like... 
> 

Also we're not testing mental aptitude but imparting knowledge.  Knowing to use flattenOn:, WriteStream, #nextPut: is 

```suggestion
Pharo's extensive class library makes the simplest solution `aCollection flattened select: #isNotNil` too trivial to gain much value from.  So the challenge is to derive your own by reviewing how `Collection>>flattened` is implemented.  

One option is to write your own `#flattenOn:` that discards `nil` before putting items onto the stream. e.g...   `
  OrderedCollection streamContents: [ :stream | 
    self flatten: aCollection onto: stream            `
You will need to differentiate between what is a collection and what is not.

An alternative is reusing `#flattenOn:` with a new `WriteStream` subclass overriding `nextPut:` to discard `nil` after they are put onto the stream.  e.g... `
  stream := MyNilFilteredWriteStream on: OrderedCollection new. 
  aCollection flattenOn: stream                 ` 
```

macta:
Ben - I don’t think we should specify solutions, the whole point is to let people derive things themselves - even if sub-optimal, they can and should explore their chosen path. For flatten - I would argue either solution is good, for some - sub-classing and overriding is evil (and that’s a fair point).

The hints have all tried to be deliberately vague so far, so we don’t give out an answer - otherwise we just do the homework for them. 

I’d be more interested in you and I having accounts with all our alternative solutions posted, and we encourage students who have found a solution that is ok for the path they have chosen, to browse those alternatives? And learn the differences.

I also worry we should be investing more time in the harder problems where there is more going on - there are quite a few later ones we haven’t touched.

I do however like the prompt for us to keep several alternatives on file, to act as an easy reference for mentors, with notes on each of them to direct a mentor to give consistent feedback (and which we hopefully might be able to automate )

bencoman:
>  sub-classing and overriding is evil (and that’s a fair point).

Sure, and thats why there are two options presented.  But also, without a prompt people will default to solving the problems the same way they've always solved them, and miss seeing some of Pharo's power.  

> otherwise we just do the homework for them

I know thats a just-a-saying but here is where we differ in philosophy - its not homework - instead, we are **teaching**.  I entirely agree with not giving away the solution, but the problems need to be "just hard enough" to be engaging.  I've toned it down to meet you half way.  Can you commit that suggestion as it is now? (btw, do you see a button for that?)  

> investing more time in the harder problems

I'll work on refactoring generation first, to help my workflow developing exercises.  Then I'll look at one of those later problems. I'd be glad if you suggest one for me to look at.

exercises/flatten-array/.meta/hints.md
macta: Coming back to this with fresh eyes - your text is quite good - and I think we should see how it works in practice (If students want to do the more direct route - I don't think we should exclude that, and then encourage them to push a bit more - but it this leads others to create a new stream, then that would be a big win. 

As an important point - this suggestion is on the wrong file (it lives in the class comment - as this one is generated). I'll submit again.

macta: Interesting that the conversation view makes your suggestion so much harder to read than the actual file change view

macta:
@bencoman in the interest of moving on - will merge this one (having incorporated your comment). We can add to it more if you feel it need more (but I think we've moved on to more interesting issues since this)

bencoman:
Yes, press on.  Further updates are only a PR away. (but the final result here is fine.)

PR #371, opened 2019-05-28
[Leap] restructure for Finder demo.

bencoman:
I was mentoring a student submission for Leap and thought this would make a good example of how to use the Finder.  Updated the class comment with a link to [this demo vidlet](https://github.com/exercism/pharo-smalltalk/raw/master/docs/images/Hint-Finder-Leap.gif) and refactored to use the found method.


macta:
This is a good idea - the second phase to doing this is to generate the Exercism templates (that come from this change). This is where a slightly better build-pipeline would be helpful (as it could do this automatically as part of the merge process) - but for now, I can do it as I have the configlet tool setup (and I don't know if it works in windows - OSX and Linux have been tested)

macta:
By the way - this highlights a useful point (and I think you are agreeing by making this change) - that the reference example is a good place for us to keep mentor notes and examples of good solutions. The latter point is interesting, as we could/should collect the best examples and include them in the Exercise@ package - partly as a reference for mentors, but equally as they may in future ease auto-mentoring where we compare AST's (or hints) of good solutions, to guide feedback.

At the moment, there is only 1 solution in the package, but we could have multiple - with some naming convention, and potentially run the tests against all of them to ensure they stay honest.

macta:
@bencoman BTW - did you want to try regenerating the template files (e.g. Ecercism | regenerate meta data) ? You need to review the readme file to get it set up (and its not really been tried on windows)

macta:
Generated associated template files based on Ben's change

PR #373, opened 2019-05-30
Rectangles solution

macta:
This is quite an interesting problem

### DIFF
--- a/dev/src/Exercise@Rectangles/LineSegment.extension.st
+++ b/dev/src/Exercise@Rectangles/LineSegment.extension.st
@@ -1,0 +1,11 @@
+Extension { #name : #LineSegment }
+
+{ #category : #'*Exercise@Rectangles' }
+LineSegment >> isHorizontal [
+	^self start x = self end x
+]
+
+{ #category : #'*Exercise@Rectangles' }
+LineSegment >> points [
+	^self isHorizontal ifTrue: [ (self start y to: self end y) collect: [ :y | (self start x)@y ] ] ifFalse: [ (self start x to: self end x) collect: [ :x | x@(self start y )]]
+]
--- a/dev/src/Exercise@Rectangles/LineSegment.extension.st
+++ b/dev/src/Exercise@Rectangles/LineSegment.extension.st
@@ -2,7 +2,7 @@ Extension { #name : #LineSegment }
 
 { #category : #'*Exercise@Rectangles' }
 LineSegment >> isRow [
-	"Answer true if the segment is on the same row"
+	"Answer true if the segment is on the same row (the internal point x@y is storing row@col)"
 	
 	^self start x = self end x
 ]
--- a/dev/src/Exercise@Rectangles/LineSegment.extension.st
+++ b/dev/src/Exercise@Rectangles/LineSegment.extension.st
@@ -1,11 +1,17 @@
 Extension { #name : #LineSegment }
 
 { #category : #'*Exercise@Rectangles' }
-LineSegment >> isHorizontal [
+LineSegment >> isRow [
+	"Answer true if the segment is on the same row"
+	
 	^self start x = self end x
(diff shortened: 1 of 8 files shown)


dev/src/Exercise@Rectangles/LineSegment.extension.st
```diff
@@ -0,0 +1,11 @@
+Extension { #name : #LineSegment }
+
+{ #category : #'*Exercise@Rectangles' }
+LineSegment >> isHorizontal [
+	^self start x = self end x
+]
+
+{ #category : #'*Exercise@Rectangles' }
+LineSegment >> points [
+	^self isHorizontal ifTrue: [ (self start y to: self end y) collect: [ :y | (self start x)@y ] ] ifFalse: [ (self start x to: self end x) collect: [ :x | x@(self start y )]]
```
bencoman: Is this meant to be all on one line?

dev/src/Exercise@Rectangles/LineSegment.extension.st
```diff
@@ -0,0 +1,11 @@
+Extension { #name : #LineSegment }
+
+{ #category : #'*Exercise@Rectangles' }
+LineSegment >> isHorizontal [
+	^self start x = self end x
```
bencoman: this doesn't look like a boolean result (??) Typo 'minus' instead of 'equals' ?

dev/src/Exercise@Rectangles/Rectangles.class.st
```diff
@@ -0,0 +1,55 @@
...
+Rectangles >> possibleRectangleRegionsIn: matrix [
+	"Answer all possible Regions in the matrix, where there could be a rectangle"
+	
+	| corners |
+	
+	corners := (matrix
+		indicesCollect: [ :r :c | 
+			(matrix at: r at: c) = $+
+				ifTrue: [ r @ c ]
+				ifFalse: [ nil ] ]) asArray reject: [ :p | p isNil ].
```
bencoman: This statement is long and hard to hold in my head.  Could it be split... `corners := corners reject: [...]` ?

bencoman:
Brute force is quite a valid approach and is the only one I could think of for a while.
Then reviewing `Array2D >> isLineAt:`  another concept came to mind...
1. Generate horizontal line segments and validate pairs as possible rectangle sides
2. Generate pairs of vertical line segments and validate pairs
3. Combinatorially compare the two collections
(not advocating you change yours, just sharing)

dev/src/Exercise@Rectangles/LineSegment.extension.st
macta: I think github doesn't show formatting properly as this is actually formatted over several lines

dev/src/Exercise@Rectangles/LineSegment.extension.st
macta: this is a boolean result - are the x values the same - however it did prompt me to change the name as horizontal ness is confusing if x@y are cartesian, where im actually using the point to represent row@column (so a bit of a stretch). I figure if the method is, isRow (and I used a comment) that makes it a bit better)

dev/src/Exercise@Rectangles/Rectangles.class.st
macta: I've refactored out a few sub methods, to make it more readable

macta:
Your comment about generate line segment permutations is exactly what this solution is doing (maybe its a bit more obvious now) - it finds square permutations (based on the + corners), and then validates if they actually exist because they are correctly joined by "-" or "|" lines. (Probably worth saying this very text in the class comment actually)

Bajger:
Solution moved and adopted for v3 in PR: https://github.com/exercism/pharo-smalltalk/pull/661

PR #381, opened 2019-06-30
[Allergies] Use plain Error

bencoman:
I'd like to use more specific  ApplicationError, but "domain" seems strongly tied to mathematics from than application when the class description is considered...
```
I am DomainError, an ArithmeticException indicating that some argument 
falls outside an expected domain, [from, to] When my valid interval is left- or right-open, 
use signal: creation protocol to provide a custom messageText rather than the default [from, to] notation.
```


Bajger (approved):

Bajger (approved):

BethanyG (approved):
@Bajger  -  seeing if approving as a member of the reviewer's group will get this to mergable.

Bajger:
@BethanyG: I don't mind to be added to reviewer's group, I'll do some more commits, reviews in future (as time allows).

PR #387, opened 2019-07-16
Reverse string streams hint

bencoman:


macta:
Might this be the cause of https://github.com/exercism/pharo-smalltalk/issues/388?

Issue #386, opened 2019-07-16
[TwelveDays] might use #joinUsing:last: 


bencoman:
A student submitted the following and its the first I knew of #joinUsing:last:
That may make a better canonical solution.

```
TwelveDays >> getVerse: anInteger [
	| elements |
	elements := (self verses first: anInteger) reverse.
	^ 'On the ' , (self dayName: anInteger)
		, ' day of Christmas my true love gave to me: '
		, (elements joinUsing: ', ' last: ', and ') , '.'
```

samWson:
I'm comparing this with `#giftsOnDay:` from the canonical solution:
```
giftsOnDay: anInteger
  ^ anInteger = 1
    ifTrue: [ self gifts first ]
    ifFalse: [
        | gifts |
			
            gifts := (self gifts copyFrom: 1 to: anInteger) reverse.
            (gifts allButLast inject: '' into: [ :str :each | str , each , ', ' ]) , 'and ', gifts last ]
```

What I like about the students solution is that returned string is close to a template. It's a bit easier to read and understand what is supposed to be returned.

I notice in the canonical solution the code critic is recommending streams over string concatenation inside iterations. I think this student solution would have the same problem. We could push for use of streams but it might make this solution less readable which is its advantage for me. At least in the case of this exercise we know the maximum number of iterations we will have.

bencoman:
I've been meaning to disable that "streams over string concatenation" critic.  Its not an absolute rule and is the one most often questioned on the Pharo mail lists.  I think there is already an issue for it.

macta:
Guys - there isn’t a “canonical “ solution per say - we just have to solve a problem to show we’ve thought about it and have a useful reference point for mentoring.

This said - updating “our solution” has been useful to help guide mentors (and I often look at them to remind me how I did it). I actually like the idea of having multiple “different” solutions in our reference baseline to reflect that there are many different ways to solve a problem (a few of our solutions have this, and I think we should do it a bit more - the trick is having a good way to run the same tests over each of them to keep them up to date).

samWson:
I see that `FlattenArray` has more than one solution. Is it a good example of how to approach having more than one solution for an exercise? If so I'd like to have a go at adding it.

macta:
Hmmm -for a moment I couldn’t find FlattenArray - but yes, that’s an example of how to do it (not the most elegant , but it does work and has the benefit that its simple to understand).

[...]

PR #372, opened 2019-05-29
Refactor ExercismGenerator to use Tonel file map

bencoman:
Working my way through the generation code, I've simplified it using Tonel file map.
`git status` after generation shows no changes except the following are deleted:
- README.md of all exercises - WIP 
- robot-simulator/.meta/solution/BearingTest.class.st - ???

PR #392, opened 2019-08-08
Add alternative solution to TwelveDays exercise

samWson:
Here is how an alternative solution to Twelve Days might look based on the solution shown in #386. The biggest change is that `#verse:` is swapped out for `#getVerse:` (might keep the name the same if we use this solution). `#verses` and `#dayName` are direct copies of the first solutions `#gifts` and `#ordinal:` respectively, the names just changed to fit in with the alternate solution. 

The big point of this solution is that it uses `SequencableCollection#>>joinUsing:last:` in `#getVerse`.

Let me know if you want to go ahead with this.

### DIFF
--- a/dev/src/Exercise@TwelveDays/TwelveDaysTest.class.st
+++ b/dev/src/Exercise@TwelveDays/TwelveDaysTest.class.st
@@ -86,7 +86,7 @@ TwelveDaysTest class >> version [
 { #category : #running }
 TwelveDaysTest >> setUp [
 	super setUp.
-	twelveDaysCalculator := TwelveDays2 new
+	twelveDaysCalculator := TwelveDays new
 ]
 
 { #category : #tests }
--- a/dev/src/Exercise@TwelveDays/TwelveDaysTest.class.st
+++ b/dev/src/Exercise@TwelveDays/TwelveDaysTest.class.st
@@ -86,7 +86,7 @@ TwelveDaysTest class >> version [
 { #category : #running }
 TwelveDaysTest >> setUp [
 	super setUp.
-	twelveDaysCalculator := TwelveDays new
+	twelveDaysCalculator := TwelveDays2 new
 ]
 
 { #category : #tests }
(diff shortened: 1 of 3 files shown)


dev/src/Exercise@TwelveDays/TwelveDaysTest.class.st
```diff
@@ -86,7 +86,7 @@ TwelveDaysTest class >> version [
 { #category : #running }
 TwelveDaysTest >> setUp [
```
macta: Shouldn't this method be in TwelveDays2Test - e.g. you are overriding the setup from the parent tests, but inheriting the actual tests? This way, the original class is no longer tested?

samWson: Odd but somehow `TwelveDays2Test>>#setup` was not a part of this commit.

samWson:
@macta will I need to generate the meta data for this additional solution i.e. as in [publishing a completed exercise](https://github.com/exercism/pharo-smalltalk#publishing-a-completed-exercise-maintainers)?

macta:
No - the exercises are only visible to mentors and if we choose to publish as if we were students ourselves.

The publishing step only effects making the tests available and we derive the readme from the test class as well.

dev/src/Exercise@TwelveDays/TwelveDaysTest.class.st
samWson: @macta are you happy with the changes? Let me know and we can get this merged.

samWson:
@bencoman I'd like to get a few of these open pull requests merged. PR's like this should not require regeneration of the tests and meta data. Will it cause problems if I merge this one (and others) before the work on #372 is merged?

bencoman:
@samWson This PR won't require regeneration.  Without refreshing myself with the generation code, I'm not sure if TwelveDays2Test would survive regeneration, but we should fix that if it does.

PR #396, opened 2019-10-07
Two extra tests for Resistor Color Duo:

glennj:
5. give 3 colors and only return the value considering the first 2.
6. give an invalid color and expect an error to be thrown.

PR #399, opened 2019-10-10
Resistor Color Duo: add new tests:

glennj:
- ensure only first two colours are used
- error if invalid colour is given

### DIFF
--- a/dev/src/Exercise@ResistorColorDuo/ResistorColorDuo.class.st
+++ b/dev/src/Exercise@ResistorColorDuo/ResistorColorDuo.class.st
@@ -19,6 +19,12 @@ ResistorColorDuo class >> resistorColorsTable [
 ResistorColorDuo >> valueWithColors: coloursCollection [
 	"Answer the value of the resister with colour bands in @coloursCollection"
 	
-	^coloursCollection inject: 0 into: [ :result :colour |
-		result * 10 + ((self class resistorColorsTable indexOf: colour) - 1) ]
+	^(coloursCollection first: 2)
+		inject: 0
+		into: [ :result :colour |
+			| idx |
+			idx := self class resistorColorsTable 
+				indexOf: colour 
+				ifAbsent: [ self error: 'Invalid color ', colour ].
+			result * 10 + (idx - 1) ]
 ]
--- a/dev/src/Exercise@ResistorColorDuo/ResistorColorDuoTest.class.st
+++ b/dev/src/Exercise@ResistorColorDuo/ResistorColorDuoTest.class.st
@@ -105,3 +105,28 @@ ResistorColorDuoTest >> test04_OrangeAndOrange [
 	result := resistorColorDuoCalculator valueWithColors: #('orange' 'orange' ) .
 	self assert: result equals: 33
 ]
+
+{ #category : #tests }
+ResistorColorDuoTest >> test05_IgnoreExtraColors [
+	| result |
+
+	result := resistorColorDuoCalculator valueWithColors: #('red' 'orange' 'yellow' ) .
+	self assert: result equals: 23
+]
+
+{ #category : #tests }
+ResistorColorDuoTest >> test06_DetectInvalidColors [
+	"first color"
+	self
+		should: [resistorColorDuoCalculator valueWithColors: #('unknown' 'black' )]
+		raise: Error
+		whoseDescriptionIncludes: 'Invalid color'
+		description: 'Should get an error with an unknown color'.
+		
+	"second color"
+	self
+		should: [resistorColorDuoCalculator valueWithColors: #('black' 'unknown' )]
+		raise: Error
+		whoseDescriptionIncludes: 'Invalid color'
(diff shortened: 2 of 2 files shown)


glennj:
Follow up PR #396

dev/src/Exercise@ResistorColorDuo/ResistorColorDuo.class.st
```diff
@@ -19,6 +19,12 @@ ResistorColorDuo class >> resistorColorsTable [
...
-	^coloursCollection inject: 0 into: [ :result :colour |
-		result * 10 + ((self class resistorColorsTable indexOf: colour) - 1) ]
+	^(coloursCollection first: 2)
+		inject: 0
+		into: [ :result :colour |
+			| idx |
+			idx := self class resistorColorsTable 
+				indexOf: colour 
+				ifAbsent: [ self error: 'Invalid color ', colour ].
+			result * 10 + (idx - 1) ]
```
glennj: As I'm new to Pharo, I don't know how the community feels about using temp vars like this. I though the command is long enough now to break it out of the `result * ...` expression.

glennj:
And I do see the new tests executed in CI: https://travis-ci.org/exercism/pharo-smalltalk/builds/596222737#L971

macta:
Hi - I'm sitting with @samWson (who is visiting London), who talked to you previously about this contribution. 

The confusion goes a bit deeper - and very sorry that this is more complicated than it should be... however let me explain (and possibly we need to update the readme.md even more on this).

Let me explain -

The tests for all exercises (and in fact all languages including Pharo, share the same definitions) come from an official exercism repo - here: https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color-duo/canonical-data.json

So any new tests that you spot, normally are contributed back to exercism (where they get discussed). Now interestingly, you have spotted that our resistor version is behind (and missing one of the similar tests to what you have proposed - so we need to regenerate to catchup). 

This said - we can add custom tests to a track - and have done so in the past, so your invalid colours test is a good one. We don't yet handle over-generation of existing tests very well (on the todo list) so the convention has been to not number them and put them in the category "extra" so we can spot them.  So I suggest we do that.

With regards to your idx (index var question) - there is no problem using temp vars (Smalltalkers prefer descriptive code - as per the refactoring book) - however as an observation, why not extract method on that expression so that you get something like:
`result * 10 + self colourIndex`

You decide whether to subsume the -1 or not.

Finally - there is another step in the readme where the exercises are generated for Exercism to process for the website. This is something we want to fix better, as submitters shouldn't need to know about it (it should generate this as part of the build process, however it currently doesn't as it needs a commit into another repo, and we need to figure the best way to do that cleanly).

Anyway for now - if we can fix the above bit, and we will also update the docs to make this easier for all of us to contribute.

glennj:
I have recategorized the extra tests, and implemented your very cood suggestion to move the messy detail into a method.

I have left the "ignore extra colours" test it: will that impede the test regeneration effort?

I did not understand your "Finally..." paragraph. Are you expecting me to perform the regeneration?

macta:
Great - I’ll get that exercise regenerated and your improvement added (I’m on holiday this weekend so stay tuned).

The finally was more an FYI and a reminder to me to sort out that generation piece.

Tim

dev/src/Exercise@ResistorColorDuo/ResistorColorDuo.class.st
```diff
@@ -15,10 +15,20 @@ ResistorColorDuo class >> resistorColorsTable [
 	^ #('black' 'brown' 'red' 'orange' 'yellow' 'green' 'blue' 'violet' 'grey' 'white')
 ]
 
+{ #category : #exercism }
+ResistorColorDuo >> colorValue: colour [
+	"Find the value of the given colour band, or error if not found."
+	
+	^ -1 + (self class resistorColorsTable
```
Bajger, line 22: Adding -1 as decrement on beginning is not very usual in Smalltalk (however it increases visibility). Maybe to assign statement into temp var and then just simply `^ temp - 1` would do same job?

dev/src/Exercise@ResistorColorDuo/ResistorColorDuoTest.class.st
```diff
@@ -105,3 +105,28 @@ ResistorColorDuoTest >> test04_OrangeAndOrange [
...
+
+{ #category : #tests }
+ResistorColorDuoTest >> test05_IgnoreExtraColors [
+	| result |
+
+	result := resistorColorDuoCalculator valueWithColors: #('green' 'brown' 'orange' ) .
+	self assert: result equals: 51
+]
+
+{ #category : #extra }
```
Bajger, line 117: what about naming category `tests-extra` ?

Bajger (changes requested):
Hi @glennj! I would treat decrement -1 in different way. Otherwise looks good! Thank you!

SleeplessByte (approved):

Issue #405, opened 2020-04-09
discord invitation for pharo needs an update


nur-azhar:
https://exercism.io/tracks/pharo-smalltalk/learning

https://discord.gg/QewZMZa

sorry i failed to figure out the file to send a pull request

iHiD:
(cc @exercism/pharo-smalltalk )

samWson:
I've updated the broken link. It should be pointing toward a functioning invitation now. Let me know if this is still an issue.

PR #406, opened 2020-04-11
Update broken Discord invite link

samWson:
The link is updated to the same as that on the http://pharo.org/community page.


samWson:
Fixes #405

Issue #407, opened 2020-04-11
CI build failing "The command 'bin/fetch-configlet' failed"


samWson:
On PR #406 the CI build is failing with the error `The command "bin/fetch-configlet" failed and exited with 2 during .`.

Not sure why this is happening yet.

samWson:
Other exercisim language tracks may have had the same problem. [Here](https://github.com/exercism/rust/pull/929) is the Rust track.

PR #408, opened 2020-04-11
Fix broken fetch-configlet script

samWson:
This seems to be a problem across many [tracks](https://github.com/exercism/configlet/issues/173). The changes here are based of those in the [Rust track](https://github.com/exercism/rust/pull/929/files). I've chosen that particular one because I can understand what the changes do.

The [original source](https://github.com/exercism/configlet/blob/master/scripts/fetch-configlet) `fetch-configlet` script has a lot of differences that are currently beyond my understanding.

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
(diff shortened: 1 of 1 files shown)


samWson:
Fixes #407 .

Issue #394, opened 2019-08-17
Sieve: alternative solution


bencoman:
A very nice solution was submitted by Exercism user "alyssap" that I find easier to understand than our current solution.  Perhaps we could adopt it...
```
Sieve >> primesLimit: anInteger [
  | primes numbers |
  primes := Stack new.
  numbers := Array new: anInteger.
  2 to: anInteger do: [ :i | 
    (numbers at: i) ifNil: [ 
       primes push: i.
       i to: anInteger by: i do: [ :j | numbers at: j put: #no ] ] ].
  ^ primes reversed
```

bencoman:
Indeed, the problem statement says "It does not use any division or remainder operation" and our current solution does use division.  Here is a slightly modified version...
```
Sieve >> primesLimit: anInteger [
  primes := Stack new.
  marks := Array new: anInteger.
  marks at: 1 put: #notPrime.
  marks doWithIndex: [ :mark :number |
    mark ifNil: [ 
      primes push: number.
      number to: anInteger by: number do: [ :notPrime | 
        marks at: notPrime put: #notPrime ] ] ].
  primes reversed
```

macta:
Where it makes sense, I’d suggest we have multiple solutions (keeping the best one of course).

I’m always impressed with the interesting solutions that keep popping up.

samWson:
@RoelofWobben would you like to take on this issue? We started collecting alternative solutions like this one an add them to our exercises.

For an example of how we have done this before checkout the [flatten array exercise](https://github.com/exercism/pharo-smalltalk/tree/master/dev/src/Exercise%40FlattenArray). You will see the `FlattenArray2` and `FlattenArray2Test` classes. Those are the extra solution for that exercise.

RoelofWobben:
Oke, 

I have the files. Do I need to generate them or can I just push my changes ?

samWson:
@RoelofWobben I don't know if the file generator would create files from alternate solutions. The flatten array exercise still only has one solution and test. In any case we still only want to have one test and solution for the website so it should be fine to not run the file generator.

macta:
Guys - the solutions are only to show that there is a solution, which unlocks the exercise - students never see them , which is why I’m always surprised at everyone being keen to add lots of alternatives... 

This said, useful alternate solutions are useful for mentors to refer to, who presumably have the dev image open with the mentor tools... so can see alternatives for reference.

[...]

PR #420, opened 2020-08-26
All the files for a second solution of the Sieve challenge

RoelofWobben:


dev/src/Exercise@Sieve/SieveTest2.class.st
```diff
@@ -0,0 +1,54 @@
+Class {
+	#name : #SieveTest2,
```
samWson, line 2: The class name should be `Sieve2Test` since it is testing the `Sieve2` class.

dev/src/Exercise@Sieve/SieveTest2.class.st
```diff
@@ -0,0 +1,54 @@
...
+	#superclass : #ExercismTest,
```
samWson, line 3: You don't need to write all the test methods like you have done here. They are exactly the same as `SieveTest` so you can reuse them with inheritance.

Make the super class `SieveTest`. You will still need to override the `#setUp` method so the instance variable `sieveCalculator` is set to a `Sieve2`. Leave `#setUp` as it is here. It is already correct.

dev/src/Exercise@Sieve/Sieve2.class.st
```diff
@@ -0,0 +1,20 @@
+Class {
+	#name : #Sieve2,
+	#superclass : #Object,
+	#category : #'Exercise@Sieve'
```
samWson, line 4: Something I forgot to tell you. We will have this in a different category. Instead of `Exercise@Sieve` put it in `Exercise@Sieve-Solution2`. That puts it in a sub category or tag.

dev/src/Exercise@Sieve/Sieve2Test.class.st
```diff
@@ -0,0 +1,11 @@
+Class {
+	#name : #Sieve2Test,
+	#superclass : #SieveTest,
+	#category : #'Exercise@Sieve'
```
samWson, line 4: Same thing here. Change the category to `Exercise@Sieve-Solution2`.

samWson:
@RoelofWobben that looks good go ahead and merge it.

samWson:
Fixes #394

PR #428, opened 2020-10-26
Remove redundant `ExercismGenerator>>#generateTo:` method

samWson:
Previous work made `ExercismGenerator>>#generateTo:` redundant. This change allows it to be safely
removed.

Storing the default path for exercises in a class variable proved to be problematic when it is mutable. In tests 
it could be overridden with a memory file system. When an attempt is then made to write exercises in production
the same memory file reference would be used. This will at the least be unintentional and possibly cause an 
error. The class variable could be reset in the test tear down but I believe the better option is to not allow 
the class variable to be mutated at all. I've removed mutation of the class variable in the tests and now 
`ExercismGenerator` has a `exercisesPath` instance variable instead of references to the class variable.

In a previous commit I made the instance variable for `exTonelWriter` lazy assigned so it could be easily
replaced with a mock in testing. However for production this did not make sense as `exTonelWriter` needed
to be initialized with a file path to write to and this will not always be known when lazy assignment takes 
place. So I have removed the lazy assignment since `exTonelWriter` will always require deliberate 
initialization.

### DIFF
--- a/dev/src/ExercismDev/ExercismGenerator.class.st
+++ b/dev/src/ExercismDev/ExercismGenerator.class.st
@@ -11,7 +11,7 @@ seperate directory suitable for the Exercism command line tool.
 
 I need two other key objects to do my job: an `ExTonelWriter` and a `PipeableOSProcess`. `ExTonelWriter` does the
 actual job of writing Tonel format files to disk. `PipeableOSProcess` is needed to make a external command line
-call to the configlet tool.
+call to the configlet tool. The configlet tool will need to be installed on the host operating system.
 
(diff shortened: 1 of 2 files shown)


samWson:
Hi @exercism/reviewers. Could I have someone review this change please? This is a bit more work to help with [exercism/v3#2264](https://github.com/exercism/v3/issues/2264).

Bajger (approved):
Hi Sam! Changes looks good to me.

samWson:
Thanks @Bajger. I'll still need one approving review from someone with write access to unblock merging. @exercism/reviewers can you help here please?

ErikSchierboom (approved):

PR #435, opened 2021-01-23
Create Exercism V3 baseline

samWson:
The intention is to create the code supporting Exercism version 3 (V3) from a clean start. Refactoring
V2 code has proven too large a task.

With a V3 baseline we can load the code and dependencies into the image without any V2 code cluttering
things up. This also allows us to keep development of both versions in the same repository.

`SourceImporter` is just a class stub so that the baseline contains something and we can verify that it
loads correctly.

The V3 baseline can be loaded using Iceberg:  In the repositories window, right click on the 
`pharo-smalltalk` repository -> mouse over `Metacello` -> left click on `Install baseline of Exercism...`
-> enter `v3` into the text field -> left click the `OK` button.

### DIFF
--- a/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
+++ b/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
@@ -51,6 +51,8 @@ BaselineOfExercism >> baseline: spec [
 						'ExercismTools' ),  
 						self class exercisePackageNames ];
 			
+				package: 'ExercismV3';
+				
 				group: 'default'
 					with: #('ExercismTools' 'Exercise@Welcome');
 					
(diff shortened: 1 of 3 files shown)


samWson:
Hi @exercism/reviewers. Could I have an approval for this PR please?

In Pharo Baselines are a tool for managing dependencies between packages in the image. They allow us to easily load groups of packages that are versioned by Git.

Having a V3 baseline makes it easy for me to load the packages separate from V2 while still using the same repository.

SleeplessByte (approved):

SleeplessByte:
@iHiD / @ErikSchierboom reviewers have no rights here

iHiD:
(You do now)

PR #467, opened 2021-03-10
Fixes #466 by specifying main branch

edward:
Fixes #466 by specifying main branch

PR #468, opened 2021-03-13
Remove outdated configlet usage from CI tasks

samWson:
CI failures on #467 raised some possible outdated CI configuration. This is to fix the incorrect configlet usage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,4 +32,3 @@ jobs:
       - run: smalltalkci -s ${{ steps.smalltalkci.outputs.smalltalk-version }}
         shell: bash
         timeout-minutes: 15
-        
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,14 +33,4 @@ jobs:
         shell: bash
         timeout-minutes: 15
         
-  tree:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v2
-
-      - name: Fetch configlet
-        uses: exercism/github-actions/configlet-ci@main
-
-      - name: Configlet Tree
-        run: configlet tree --with-difficulty .
+        
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
         uses: exercism/github-actions/configlet-ci@main
 
       - name: Configlet Linter
-        run: configlet lint --track-id pharo .
+        run: configlet lint .
     
   test:
     name: Test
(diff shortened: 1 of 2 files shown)


samWson:
The `tags` array in `config.json` is required to be non-empty by `config lint`. It's been updated with the relevant tags.

samWson:
Hi @exercism/reviewers. Can I please get an approval for this PR? Anyone with `configlet` and Github Action experience would be ideal.

.github/workflows/ci.yml
```diff
@@ -33,14 +33,4 @@ jobs:
...
-
-    steps:
-      - uses: actions/checkout@v2
-
-      - name: Fetch configlet
-        uses: exercism/github-actions/configlet-ci@main
-
-      - name: Configlet Tree
-        run: configlet tree --with-difficulty .
+
```
ee7, line 36:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -33,14 +33,4 @@ jobs:
         shell: bash
         timeout-minutes: 15
```
ee7, line 35:
```suggestion
```

ee7 (commented):
Hi.

The track now has a dedicated file at [.github/workflows/configlet.yml](https://github.com/exercism/pharo-smalltalk/blob/main/.github/workflows/configlet.yml) that runs configlet:

https://github.com/exercism/pharo-smalltalk/blob/37cb9b1b940cd0092955338a7a59c0671bf6a8de/.github/workflows/configlet.yml#L1-L16

So we no longer need to run configlet in `ci.yml` at all - it's better to remove lines 9-20 inclusive in [the current latest commit of this PR](https://github.com/exercism/pharo-smalltalk/blob/7bfec4bc08b9b5814134000ce6c6a1ca8b3106ac/.github/workflows/ci.yml)

And at some point (in a later PR if you prefer), we should also rename `master` to `main` in `ci.yml`. Currently, `ci.yml` doesn't run on push to the main branch - see [the run for the latest commit to main](https://github.com/exercism/pharo-smalltalk/runs/1887536372) where only the `configlet.yml` workflow ran.

Otherwise, looks good - `configlet` failed before the PR, and passes with the PR.

I didn't review whether the tags make sense for pharo-smalltalk.

I don't have write access to this repo, so an approving review from me doesn't allow merging anyway.

samWson:
@ee7 Thanks for the review. I've made some changes as suggested.

ee7 (approved):
You're welcome.

Approved, as long as you're aware that the `ci.yml` workflow won't actually run when a PR is merged - at least until `master` is changed to `main` here: 

https://github.com/exercism/pharo-smalltalk/blob/37cb9b1b940cd0092955338a7a59c0671bf6a8de/.github/workflows/ci.yml#L1-L6

It looks like #467 makes some other `master` -> `main` fixes, but not this one.

iHiD (approved):

SleeplessByte (approved):
You might want to create TAGS.md. I did that for JavaScript explaining why we chose the tags we chose. 

Just for future people looking at it 😬🤗

samWson:
@ee7 Thanks. I missed the branch name and have changed it to `main`. @iHiD and @SleeplessByte thanks for the approvals. I'll have a look at the JavaScript `TAGS.md` and make a similar one.

Issue #506, opened 2021-09-21
Fix link to track progress


RoelofWobben:
when you now click on this link

![afbeelding](https://user-images.githubusercontent.com/89863657/134216093-12087201-f6da-4958-8689-03cf413eb2b5.png)

you get a 404.

The link on the normal image and the dev image needs to point to https://exercism.org/tracks/pharo-smalltalk

RoelofWobben:
This one can be solved here

![afbeelding](https://user-images.githubusercontent.com/89863657/134218972-1fe6f968-d60a-4326-a82d-f852798065e5.png)

RoelofWobben:
somewhere a hint where  I can find the ExercismManager.
Then this is one of the easist to make PR.

RoelofWobben:
Is solved by PR 507

PR #507, opened 2021-09-21
Fix menu choice  see track progress

RoelofWobben:
Solves the menu choice  where you can look your track progress

solves : #506


dev/src/ExercismTools/ExercismManager.class.st
```diff
@@ -197,5 +197,5 @@ ExercismManager >> submitToExercism: packageOrTag [
 
 { #category : #exercism }
 ExercismManager >> viewOnExercism: packageOrTag [
-	WebBrowser openOn: 'https://exercism.io/my/tracks/pharo'
+	WebBrowser openOn: 'https://exercism.org/tracks/pharo'
```
gypsydave5, line 200: This is the only change we want

RoelofWobben: I know , made a new fork and new changes and still all these unrelated things show up.

RoelofWobben:
Close this because a lot of unrelated commits are in it

PR #511, opened 2021-09-22
fix the generic get test

gypsydave5:
broken as it was asserting on something that wasn't in the current Google response

Frankly these tests shouldn't really exist; suggest injecting a ZnClient into an ExercismClient and then mocking behaviours we care about


Bajger (approved):

Bajger (approved):

RoelofWobben:
@ErikSchierboom  @Ihid can someone approve this and merge.

ErikSchierboom (approved):

PR #516, opened 2021-10-18
Proposing Exercism v3 concepts description

Bajger:
- Added MD file with initial work-in-progress description of concept exercises
- purpose of this summary is to review chapters (concepts) and start working on initial concept exercises (individual MDs per exercise)
- for each exercise concept produce comprehensive content based on brief summary in this file


RoelofWobben:
oke, but I still do not know if that is what the exercism maintainers wanted. 
As far as I know and see  they would see a very small file with explanation of 1 concept.
And it looks to me, the user gets a lot of info.

Bajger:
> As I see it now. the first 3 "chapters" could be some docs on the page

Why do you think so? Maybe first is just intro (which is also seen at other tracks). But Pharo object model is something has some practical swowcases, as well as syntactic elements on a postcard. It isn't just documentation or remote links. But it doesn't really need to be discussed at the moment, but rather on concrete concept exercise. Just see MOOC how it is approached. To me, each chapter is important and has practical goals / objectives.

RoelofWobben:
oke, then im curious to see what challenge can be at for example the pharo object model
Most tracks I see uses the lasanga challenge as first one after the helllo - world

PR #557, opened 2023-01-10
Update contributing guidelines

### LINKED ISSUE #556: Update conributing guidelines to be up-to-date
Adding some contextual information and details that might be useful for other contributors who wants to work on bugfix or exercise:
- restructured guidelines by knowledge part, git repositories, Pharo image and CLI tools
- added overview of how to work on exercise
- added or rewritten some details when completing/publishing exercise.

Bajger:
- fixes #556


Bajger:
@ErikSchierboom Hello Erik! I'm bit unsure about using `bin/configlet sync` command. Is it needed to produce file artefacts when creating new from problem specification? I didn't mention this in contributing guidelines so far. Thanks for reply!

ErikSchierboom:
> Is it needed to produce file artefacts when creating new from problem specification? I didn't mention this in contributing guidelines so far. Thanks for reply!

Well, it is not needed per se (you could manually copy the data), but I would recommend using `configlet sync`.

Bajger:
I've updated the instructions with up-to-date information. So potential contributors can use it in future. I'm aware some of the old code, when I wrote instructions. (e.g. `configlet generate` is not needed for practise exercise generation - so it needs to be removed, instead of  it a `sync` command should be invoked, when exercise is generated from Pharo image).

ErikSchierboom (approved):

PR #577, opened 2023-05-23
Matrix exercise: change solution and example filenames

glennj:
The test generator creates a `MatrixReader` class, but the exercise config uses a `Matrix` filename.

http://forum.exercism.org/t/test-runner-fails-on-matrix-exercise/4346


Bajger (changes requested):
I think for convenience, test class should be renamed too (from MatrixTest to MatrixReaderTest). Test uses proper solution class name already - MatrixReader.
See: https://github.com/glennj/exercism-pharo-smalltalk/blob/matrix/dev/src/Exercise%40Matrix/MatrixTest.class.st

**Update**:
I'm sorry for contradictionary comment, maybe more better is to rename solution class name and change it in the test. Because exercise is called "Matrix" and not "Matrix reader". 
See: https://github.com/glennj/exercism-pharo-smalltalk/blob/30d9f118cde8f4d6dab897a1ad8ba68a92aa26da/dev/src/Exercise%40Matrix/MatrixTest.class.st#L94
https://github.com/exercism/problem-specifications/blob/main/exercises/matrix/metadata.toml

dev/src/Exercise@Matrix/MatrixTest.class.st
```diff
@@ -71,7 +71,7 @@ MatrixTest class >> exercise [
 MatrixTest class >> generator [
 	"Potential generator hints"
 
-	^ {('@Matrix' -> '#MatrixReader').
+	^ {('@Matrix' -> '#Matrix').
```
glennj, line 74: Is this still needed? I don't know the meaning of `@` and `#`

exercises/practice/matrix/MatrixTest.class.st
```diff
@@ -71,7 +71,7 @@ MatrixTest class >> exercise [
 MatrixTest class >> generator [
 	"Potential generator hints"
 
-	^ {('@Matrix' -> '#MatrixReader').
+	^ {('@Matrix' -> '#Matrix').
```
glennj, line 74: Still needed?

dev/src/Exercise@Matrix/MatrixTest.class.st
Bajger: It seems to be utilized by Trait called "TExercise" that should generate new v3 exercise and all its metadata. Trait is not used at the moment (no references to it) and honestly I'm just guessing intention of original author. The `generator` method seems to serve as allias for solution class and package name that should be generated into directory with vanilla/example solution. 
TLDR: You can safely remove this method from this class.

Reasoning: It looks for package name with string suffix `@Matrix` and  generator will use then solution class name with symbol `#Matrix` (symbols are just unique strings compiled in code). We don't need that conversion, package name 'Exercise@Matrix' corresponds to solution class `Matrix` (after your changes renaming class name and in usage in test).

Bajger (changes requested):
Please remove `generator` method from both places

Bajger (approved):

Bajger:
@glennj Thanks you!

PR #578, opened 2023-05-29
Change import of sources to image so that example solutions are removed

### LINKED ISSUE #576: Test runner passes unexpected on empty class definition with no method defined
**Reported by community member:** 
image

**Root cause of unexpected pass of test:**
TestRunner Pharo image contains by definition all its example (vanilla) solutions. When empty solution class that comes from Online editor is loaded into image, it doesn't overwrite originally compiled example methods with solution. 

**Solution**:
Example solution classes have to be removed prior version of class (coming from online editor) is loaded into image.

Bajger:
Change import of sources to image so that example solutions are removed prior import of student solution

- added unit test to check reming solution classes
- added example (golden) test to test empty class definition with no method defined and example present in image
- moved example (golden) tests to separate packages
- fixes #576

### DIFF
--- a/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
+++ b/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
@@ -1,3 +1,6 @@
+"
+Project baseline to load entire Exercism Pharo track, including all its dependencies
+"
 Class {
 	#name : #BaselineOfExercism,
 	#superclass : #BaselineOf,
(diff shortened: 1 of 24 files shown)


Bajger:
@glennj Thanks for taking a look. Changes consists of:
-  removing example solution, before student solution is loaded into image (as reported in related issue)
- embraced "golden" tests for test runner for cases that weren't covered before 
- moved golden tests to indenpendent packages for each golden test case

ErikSchierboom (approved):

PR #583, opened 2023-09-07
Fix exercise test generator

Bajger:
Following changes make Exercise test generator working again. Previously it resulted in error and therefore test classes cannot be generated from canonical descriptions.


Bajger:
@glennj This is prerequisite/fix to generate test classes out of canonical problem specifications. After that I can generate up to date test cases in Pharo, which would test solution for exercise. Thanks for review!

ErikSchierboom (approved):

glennj (approved):

glennj:
Thanks @Bajger !

PR #584, opened 2023-09-08
Add circular buffer exercise

Bajger:
Added circular buffer exercise including solution.


Bajger:
@glennj I have some trouble with configlet lint, with UUID - obviously Pharo generator (by: `UUIDGenerator next`) does not produce in format that is needed for linter. Will find out, how it can be fixed (worst case I will overwrite it manually).
Other than that, you can review tests and solution for circular buffer. Hope it is Pharo idiomatic way.

dev/src/Exercise@CircularBuffer/CircularBuffer.class.st
```diff
@@ -0,0 +1,92 @@
...
+]
+
+{ #category : #operations }
+CircularBuffer >> overwrite: aValue [
+
+	^ self bufferFull
+		  ifTrue: [
+			  self buffer removeFirst.
+			  self buffer addLast: aValue ]
+		  ifFalse: [ self write: aValue ]
```
glennj, line 64: re-implementing `write` in here. How about
```smalltalk
self bufferFull ifTrue: [ self buffer removeFirst ].
^ self write: aValue
```

glennj (approved):

Bajger:
CI green! @ErikSchierboom can approve now.

glennj (approved):
Looks great!

ErikSchierboom (approved):
Great!

PR #586, opened 2023-09-13
Refactoring of ExercismExerciseGenerator

Bajger:
Generation of test class, its cases from problem specifications.
- added pragmas with test case metadata
- otherwise result should be same as original implementation


Bajger:
@glennj : I've refactored exercise test generator that is used when building new exercise from canonical description. Result of this generator is Test class in Pharo that contains test case methods and setup (placed in WIP package). Thanks for reviewing this!

Bajger:
@glennj @ErikSchierboom Thanks for approving this! Changes should be save (not used in usual user production code, only when generating new exercise) to use.

glennj (approved):
go for it!

ErikSchierboom (approved):

PR #611, opened 2024-02-22
610 all pharo tests pass even though i didnt solve the exercise

### LINKED ISSUE #610: All pharo tests pass even though i didn't solve the exercise
After failing to find my way around the pharo-vm locally, I decided to try to work in the browser. The hello world, exercise was easy enough, but without knowing smalltalk, I didn't know how to start the Roman Numerals exercise, or even what the class should like or what a pass or a fail would look like. I was faced with an empty page, no skeleton of a class written for me like I was expecting.

So I copy pasted the hello world class into the roman numerals exercise. I was hoping that the test would fail and I would learn what my class should be named and I'd get an error message that I could google. Instead, all tests were green:

https://exercism.org/tracks/pharo-smalltalk/exercises/roman-numerals/solutions/youve

Bajger:
- fixes #610

### DIFF
--- a/dev/src/ExercismPharo110/Class.extension.st
+++ b/dev/src/ExercismPharo110/Class.extension.st
@@ -1,6 +1,6 @@
 Extension { #name : #Class }
 
-{ #category : #'*ExercismTools' }
+{ #category : #'*ExercismPharo110' }
 Class >> removeFromSystem: logged [
 	"Forget the receiver from the Smalltalk global dictionary. Any existing  instances will refer to an obsolete version of the receiver.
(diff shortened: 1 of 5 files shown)


dev/src/ExercismPharo110/Class.extension.st
Bajger, line 1: As per discussion on Pharo Discord. Solution recommended by one of founders: https://discord.com/channels/223421264751099906/278558427796602882/1209857931022049280

ErikSchierboom (approved):

glennj:
I don't think my review would have added any value. Kudos for the fix.

PR #619, opened 2024-04-02
Fixed dependency in project baseline to be loaded in P11

### LINKED ISSUE #610: All pharo tests pass even though i didn't solve the exercise

Bajger:
I used wrong package for dependency therefore fix was not applied / loaded for Pharo 11
- fixes #610

### DIFF
--- a/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
+++ b/dev/src/BaselineOfExercism/BaselineOfExercism.class.st
@@ -70,7 +70,7 @@ BaselineOfExercism >> baseline: spec [
 			].
 			spec for: #'pharo11.x' do: [
 				"ExercismTools methods will add methods from compatibility package for Pharo 11"
-				spec package: 'ExercismTests' with: [ spec includes: #('ExercismPharo110') ].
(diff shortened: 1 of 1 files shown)


Bajger:
@glennj I'm sorry - issues was still not fixed, I used accidentally wrong package to load dependency. This should fix this. Thanks for review!

Bajger:
@ErikSchierboom: You can approve! Thanks!

glennj (approved):
woopsie!

ErikSchierboom (approved):

PR #625, opened 2024-05-06
Fix typos in `CircularBufferTest.class.st`

citizen428:
I noticed two typos in this test file.

### DIFF
--- a/dev/src/Exercise@CircularBuffer/CircularBufferTest.class.st
+++ b/dev/src/Exercise@CircularBuffer/CircularBufferTest.class.st
@@ -45,7 +45,7 @@ Because there is space available, if the client again uses overwrite to store C 
     [C][D][7][8][9][A][B]
 
 ## Hint
-For error cases during operations, see Error class hiearchy (class side) to find out how to signal exception with given message.
+For error cases during operations, see Error class hierarchy (class side) to find out how to signal exception with given message.
 
 "
 Class {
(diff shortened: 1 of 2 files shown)


dev/src/Exercise@CircularBuffer/CircularBufferTest.class.st
```diff
@@ -144,7 +144,7 @@ CircularBufferTest >> test05_FullBufferCantBeWrittenTo [
 
 	<exeTestName: 'full buffer can''t be written to'>
 	<exeTestUUID: '2af22046-3e44-4235-bfe6-05ba60439d38'>
-
```
citizen428, line 147: All these whitespace changes where just auto-fixed by my editor, I can undo those if needed.

citizen428:
Yup, my bad, fixed.

Bajger (changes requested):
Hi! Actually, both places need to be changed. 
[dev/src/Exercise@CircularBuffer/CircularBufferTest.class.st](https://github.com/exercism/pharo-smalltalk/pull/625/files#diff-9387f24f0e801f88080d2c47a149087759524e7dff9a34569f204465aefd4a96)
and that you've originally changed: https://github.com/exercism/pharo-smalltalk/blob/main/exercises/practice/circular-buffer/CircularBufferTest.class.st

Usually I change first and then run generation to exercise directory, but you don't need to do that for small change. Both files should be in sync. Sorry for confusion.

citizen428:
All good, done.

Bajger (approved):

Bajger:
@citizen428 Thank you!

PR #631, opened 2024-05-23
Fixed removal of exercise sources befor running test runner.

### LINKED ISSUE #630: Test runner fails on Minesweeper exercise
Test runner fails when minesweeper tests are executed from online editor:
https://forum.exercism.org/t/minesweeper-48in24-exercise/11292/3

**Current situation**
When running tests from Online editor on Minesweeper, either there is unexpected error, or timeout error.

**Expected**
Test runner should work on Minesweeper. 

**Root cause**
Minesweeper vanilla solution Pharo package contains defined classes `Minesweeper` as well as `MinesweeperTest`. 
Package also contains additional extension methods on `Array`, `Character` and `Integer` classes. 
Problem is that test runner removes vanilla solutions (prior student solution import) using `ExercismExercise>>removeSolutionClasses`. This causes the trouble, since underlying method `solutionClasses` gives also system classes (those which define extension methods). Method then tries to delete system classes, which leads to image destruction (e.g. removing Array class). 

**Proposed solution**
Removing sources shold remove only defined classes in exercise package as well as extension methods on system classes (but not classes itself)

Bajger:
- fixes #630

### DIFF
--- a/dev/src/ExercismTestRunner/ExercismTestRunner.class.st
+++ b/dev/src/ExercismTestRunner/ExercismTestRunner.class.st
@@ -71,8 +71,8 @@ ExercismTestRunner >> runTestsOnExercise: slugName importSourceFrom: inputDirect
 	[	
 		exercise := self findExercise: slugName.
 		
-		"remove all example solutions - solution classes prior student solution import"
-		exercise removeSolutionClasses.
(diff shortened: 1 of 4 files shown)


Bajger:
@glennj Thanks for reviewing this! @ErikSchierboom thanks for approval!

ErikSchierboom (approved):

PR #646, opened 2025-02-24
Update darts to v3

Bajger:


exercises/practice/darts/.meta/hints.md
BNAndras, line 1: Pythagorean theorem not theorum in the append

BNAndras (approved):

Issue #681, opened 2026-09-06
Fetching Exercises via `ExercismTools` Is Blocked by Cloudflare's Challenge


mckzm:
### Summary

Trying to fetch exercises using `ExercismTools` in a Pharo 12 image (same thing for 10 or 11), e.g. by using the `Fetch new exercise...` option of the `Exercism` menu, results in an Exception being thrown (`TonelParseError: Can't parse type.`) due to the request being gated by Cloudflare's challenge. The `ZnClient` doesn't get past the challenge: instead it retrieves the HTML of the challenge page, that the TonelParser errors out on.

#### Workaround
Set the `User-Agent` value of the `ZnClient` to that of the CLI Tool when instantiating it in `ExercismHttpClient >> createZnClient`, e.g.:

```smalltalk
createZnClient
    "Override User-Agent header to supply the CLI's, otherwise Cloudflare's challenge breaks us"
	^ ZnClient new
		https;
		host: self hostName;
		headerAt: 'User-Agent' put: 'github.com/exercism/cli v3.5.8 (darwin/arm64)';
		yourself
```

### Reproducer

1) Follow the [Installing Pharo locally](https://exercism.org/docs/tracks/pharo-smalltalk/installation) instructions to set up a new Pharo environment with `ExercismTools`
  * I used `$ curl https://get.pharo.org/120+vm | bash` to get a Zeroconf Pharo 12 image as that seems to be the most recent version that's supported by `ExercismTools`; the same issue exists with 10 and 11 though
  * validate `WelcomeTest>>testWelcome` passes
2) Attempt to fetch `hello-world` via the top menu `Exercism -> fetch new exercise...` option
  * fill in API token when prompted
3) an error window appears: `TonelParseError signal: 'Can''t parse type.'`

#### TonelParseError

<img width="1727" height="1083" alt="Image" src="https://github.com/user-attachments/assets/98238b73-29b7-4b10-8fee-4a9f2328c472" />

#### HTML `contentData`

<img width="1727" height="1083" alt="Image" src="https://github.com/user-attachments/assets/9195ea66-bf06-4f20-870a-f062b8063868" />

#### Relevant Excerpts from `PharoDebug.log`

(in a `Details` section as they're longish)

<details>

```zsh
$ pharo-smalltalk: bat -pP PharoDebug.log
THERE_BE_DRAGONS_HERE
Can't parse type.
6 September 2026 11:13:05.794926 am

[...]

PR #682, opened 2026-09-20
681 follow-up: align solution path and track API versioning and add User-Agent header to ZnClient

mckzm:
Refs https://github.com/exercism/pharo-smalltalk/issues/681

* Aligns the Solution path API version on the track's
* Adds a method to set a `pharo-smalltalk`-specific User-Agent header on ZnClient instances


dev/src/ExercismTools/ExercismHttpClient.class.st
```diff
@@ -79,7 +79,7 @@ ExercismHttpClient class >> reset [
 
 { #category : 'accessing' }
 ExercismHttpClient class >> solutionPath [
-    ^'/v1/solutions'
+    ^ '/v{1}/solutions' format: { ExercismManager trackVersionString }
```
glennj: I think this is wrong. `v1` here is the Exercism API version, not the Pharo track version.

https://github.com/exercism/website/blob/main/config/routes/api.rb -- lines 5 and 9

```suggestion
    ^ '/v1/solutions'
```

mckzm: Thanks for this!! 🙇 The change broke exercise fetching, which I had failed to notice 🤦‍♀️ Suggestion applied in https://github.com/exercism/pharo-smalltalk/pull/682/commits/8b507ed043365ecaf7c1960ccce2f6d32ce34382

dev/src/ExercismTools/ExercismHttpClient.class.st
```diff
@@ -275,6 +276,15 @@ ExercismHttpClient >> solutionPathFor: aString [
 						ifFalse: [ aString ]) ]
 ]
 
+{ #category : 'options' }
+ExercismHttpClient >> userAgentString [
```
Bajger: Since most of the 'constant-like' methods are put to class side, I would assume this method should go to class side as well. It contains hardcoded strings with repository name which is not object instance specific.

dev/src/ExercismTools/ExercismHttpClient.class.st
```diff
@@ -115,6 +115,7 @@ ExercismHttpClient >> createZnClient [
 	^ ZnClient new
 		https;
 		host: self hostName;
+		headerAt: 'User-Agent' put: self userAgentString;
```
Bajger: `self class userAgentString` - see comment below.

Bajger (changes requested):
@mckzm Thanks for your effort! Minor improvement suggested as per my comments.

mckzm:
Hi @Bajger :wave: thanks for your review and comments!! I addressed both in commit https://github.com/exercism/pharo-smalltalk/pull/682/commits/5dfd0ad337a8234836a185d69a22fd0583bc76cd

dev/src/ExercismTools/ExercismHttpClient.class.st
mckzm: Addressed in https://github.com/exercism/pharo-smalltalk/pull/682/commits/5dfd0ad337a8234836a185d69a22fd0583bc76cd

dev/src/ExercismTools/ExercismHttpClient.class.st
mckzm: Addressed in https://github.com/exercism/pharo-smalltalk/pull/682/commits/5dfd0ad337a8234836a185d69a22fd0583bc76cd

Bajger (approved):

mckzm:
@Bajger I pushed another commit, https://github.com/exercism/pharo-smalltalk/pull/682/commits/3c2de038b8d52efa45b023768591de3098e94e69, to fix [the failing test run](https://github.com/exercism/pharo-smalltalk/actions/runs/35707023720/job/106685590779)

Bajger:
@mckzm Thanks for your PR! This is done ✅

